PR #5, opened 2019-09-11
Feature: Create Action 

zacck:
## Changes Proposed 

Build a form for creating new community actions 

### Work 
- [x] Build form 
- [x] Validate form
- [x] Make form responsive for various screens 
- [x] Translate and localize all text assets
- [x] Submit Action to Blockchain
- [x] Display Action Creation Result


zacck:
Hey @lucca65  I habe a few remarks after taking some time to go through the comments from the other PR 

#### Comments
- I am investigating the sizing of the labels and how to get a uniform predictable result across multiple screen sizes 
- About the urls true `community/BES/objective/1/action/new` does read more semantically, will consider going for it however I chose `new/action/BES/10` for it's terseness
- All the labels and tooltips will be localised 
- I will consider componetizing the form inputs, I just hadn't gotten round to it 

#### Queries
- At the moment setting the new font default will mean all the other places depending on the older setting for Roboto will now reflect the new font family, It's ok to do I am just not sure on if I should do it **(should we do it or not?)**
- I am confused as to how I can use this library  https://package.elm-lang.org/packages/cuducos/elm-format-number/latest/ to mask the date inputs as it seems to be a library meant for floats **(Help me see)**

lucca65:
I’ve just asked Helton about this font thing and the new design font fallback will be the same as the old one, only change is that the first on the list is different: old design uses Roboto new one uses Gotham Rounded

so we can just copy the old fallbacks for the new css and leave the old one as is

zacck:
## Display Validation errors 

Screen Shot 2019-09-13 at 10 00 58

zacck:
Hey @lucca65  @alissonfpmorais  here is the MVP for the Action submission form so sorry for taking so long but I had to try make sure all the details are covered.

I am facing a few small issues 

1. The authentication modal doesn't want to dismiss after the user has autheticated with my current strategy 
2. Event source is falling over whenever I try submitting actions
If anyone has any ideas why this could be occuring I would appreciate the help

Taking a few minutes for lunch and will be back

lucca65:
@alissonfpmorais I'll be merging this so we can move forward, but please feel free to add your comments

PR #21, opened 2019-10-02
Bugs and fixes

lucca65:
- [x] Add LaCoopmunitat logo
- [x] Adjust aspect ratio of logos in activity panel
- [x] Adjust aspect ratio of logos in new action form
- [x] Make claim button work on mobile
- [x] Send EOS errors to Sentry
- [x] Fix Eye icon during claim, using absolute path instead of relative
- [x] Make Dashboard 'last actions' empty (it was a showing transfers instead of actions)
- [x] Fix profile update bug (adding a bio would be ignored and used the email instead)


src/elm/Page/Dashboard.elm
```diff
@@ -287,24 +287,7 @@ viewSections loggedIn model =
                 Page.viewCardEmpty [ text (t "community.actions.no_actions_yet") ]
 
             LoadedGraphql transfers ->
```
zacck: @lucca65  perhaps the naming here should be changed to actions ...

src/index.js
```diff
@@ -632,6 +632,14 @@ async function handleJavascriptPort (arg) {
             error: error
           }
           devLog('eos.transaction.failed', errorResponse)
+          // Send to sentry
```
zacck: :+1

src/elm/Page/Dashboard.elm
lucca65: Hey @zacck this line wasn't changed, I've just let it as actions. To make this `actions` we actually have to create a new query, we'll have to discuss this in more depth first. I would say that this is out of the scope for this PR

lucca65:
@zacck I'll merge this for now!

PR #24, opened 2019-10-03
Feature: Add visual cue for unread notification

zacck:
## Changes Proposed

- [x] Add visual for unread notification 
- [x] Mark notification as read when notification is clicked (Open app and handle from app)
- [x] Modify model to have unread notifications 
- [x] Fetch number of unread notifications from server 
- [ ] Highlight unread notifications from server (Handle from app)

### DIFF
--- a/public/service-worker.js
+++ b/public/service-worker.js
@@ -60,8 +60,18 @@ self.addEventListener('push', function (event) {
 // Handle Notification Clicks
 self.addEventListener('notificationclick', function (event) {
   const notification = event.notification
-
   notification.close()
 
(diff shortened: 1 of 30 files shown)


zacck:
@lucca65  so the marking of the unread notifications happens in the list that you are building using the union and as for the marking of the notification as read I need to modify the server to send the history id as well as the notification data .... 

However the issue arises when I need to handle the click on the notification, this is because importing node modules into service workers is strenous hence will take a bit of a while

zacck:
@lucca65  would you mind reviewing this PR for me please?

PR #30, opened 2019-10-24
Fix/invited reward default

lucca65:
Bug appeared in La Coopmunitat


lucca65:
Hey Zacck those commits are just master merges, since we don't actually commit anything to the master. Don't worry they won't appear on our history, since we do squash before merging!

PR #33, opened 2019-10-31
Feature: community actions

zacck:
## Changes Proposed 

- [x] Refactor claims query to Community namespace 
- [x] Add fetch claims to community page 
- [x] Display Claimed Actions on Community Page


src/elm/Page/Community.elm
```diff
@@ -83,6 +126,7 @@ type alias Model =
     , openObjective : Maybe Int
     , modalStatus : ModalStatus
     , messageStatus : MessageStatus
+    , actions : Maybe ActionVerificationsResponse
```
lucca65: Instead of using maybe we can use a type to allow loading, what do you think?

zacck: @lucca65 sure I considered that but I felt to avoid it since the types in this modules are so much kinda needs a rework, Our codebase has lots of standards didn't want to add more confusion

lucca65: sure thing bro, the PR is approved, please merge if you feel it is ready

lucca65: actually I'll go ahead and merge it!

zacck: Thanks I was fast asleep

PR #86, opened 2019-11-26
   Fix: autocomplete text

### LINKED ISSUE #76: Over imposed text
**Describe the bug**
Over imposed text when searching for a user to do a transfer. [iphone]
Over imposed numbers in PIN [Dell Computer]

**Expected behavior**
over imposed text

**Screenshots**
WhatsApp Image 2019-11-14 at 16 48 26

WhatsApp Image 2019-11-14 at 17 00 02

borboins.cambiatus.io

 - OS: Windows
 - Browser Chrome
 - Version 

 - Device: iPhone 7
 - OS: 
 - Browser  Safari
 - Version

zacck:
## Changes Proposed 

Fixes #76 

- [x] Adjust user autocomplete styles to prevent corruption when displaying 

## Screens  
Screenshot 2019-11-26 at 09 13 08

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "frontend",
-  "version": "1.5.4",
+  "version": "1.5.5",
   "private": true,
   "engines": {
     "node": "11.15.0"
--- a/src/elm/Page/Dashboard/Community.elm
+++ b/src/elm/Page/Dashboard/Community.elm
@@ -425,8 +425,8 @@ viewAutoCompleteItem shared profile =
             shared.endpoints.ipfs
     in
     div [ class "autocomplete-item" ]
-        [ Avatar.view ipfsUrl profile.avatar "profile-img-avatar-select"
-        , p [ class "whitespace-no-wrap overflow-hidden" ]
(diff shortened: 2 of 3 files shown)


lucca65:
also the tailwind configurations aren’t something i like. the ideia for the configurations is to have our style over helton’s design. it has to be consistent. having super exact values such as px don’t look good. old screens should live in the legacy_main.css and new design screens on tailwind

zacck:
@lucca65  so would you rather I use the old css files and just modify those? Then when a new design  for auto complete items comes along we can refactor to tailwind?

lucca65:
yeah! that is what i was thinking, so we don’t mess up tailwind with old stuff! makes sense?

zacck:
@lucca65  I like that plan!

zacck:
lol @lucca65  your suggestion ended up working much better!!!!

Screenshot 2019-11-27 at 07 56 28

lucca65:
@zacck before merging can you change your user name to something big with 256 chars? that's the max allowed and it should work fine with long surnames

zacck:
@lucca65  cool let me do that

zacck:
@lucca65  here is how it looks now 
Screenshot 2019-11-28 at 07 34 30

PR #109, opened 2019-12-10
Feature: Start Narrating what our code is doing 

zacck:
## Changes Proposed 

- [x] Document UpdateResult 
- [x] Document Main module

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "frontend",
-  "version": "1.7.0",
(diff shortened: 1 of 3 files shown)


zacck:
@lucca65  what do you think, can I explain this better?  Are the explanations accurate 
as for the Main module it seemed like a few comments would suffice as the module kind of explains itself?

zacck:
Will write much more docs as we keep going

PR #112, opened 2019-12-16
Enhancement/tab layout

### LINKED ISSUE #104: Change select element to tab element in Shop
To make it easy for the user to find your offers. Also removing the filter "Offers from my communities", leaving only "All offers" (from the community I'm accessing) and "My offers".

**DESK**
Captura de Tela 2019-12-09 às 10 26 10

**MOBILE**
Captura de Tela 2019-12-09 às 10 34 30

**interface attributes**
https://www.figma.com/file/GyRvSEQNS91DOLgMIk3u77/shop-desk?node-id=99%3A0

and

https://www.figma.com/file/X0ROabI93lbH0YsHDiwcFf/shop-mobile?node-id=1%3A1109

JerryNyoike:
Closes #104


zacck:
@JerryNyoike  you should add a `Closes ..` tag to connect this to an issue

src/elm/Page/Shop.elm
```diff
@@ -553,6 +553,36 @@ viewShopFilterDropdown translations maybeFilter loggedIn =
     viewMenuFilterDropdown ClickedFilter decoder options
 
 
+viewShopFilterTab :  ( String, String, String ) -> Maybe Filter -> LoggedIn.Model -> Html Msg
```
zacck: @JerryNyoike  how come a Maybe Filter, if its always tabs shouldn't a filter always be provided, How could it end up being a Nothing?

JerryNyoike: Oh, I see. I was looking at it from the point of view that only one filter could be applied at a time.I will try it this way instead

src/elm/Page/Shop.elm
```diff
@@ -553,6 +553,36 @@ viewShopFilterDropdown translations maybeFilter loggedIn =
...
+viewShopFilterTab translations maybeFilter loggedIn =
+    let
+        ( communities, all, user ) =
+            translations
+
+        decoder =
```
zacck: @JerryNyoike  I know that is what existed before but I suspect the issue is arising somewhere here, I dont see why you need to go through translations for the tab is it possible to get rid  of this and use a Custom Type instead to see which tab is actually active?

Also there is left over code for the dropdowns perhaps you can remove that too?

src/styles/legacy_main.css
```diff
@@ -3961,3 +3961,60 @@ input[type=number]::-webkit-outer-spin-button {
     -webkit-appearance: none;
     margin: 0;
 }
+
```
zacck: @JerryNyoike  hey you need to use tailwindcss to do all the styling in general we don't grow the `legacy_main.css` file check out how to use tailwind at https://tailwindcss.com/

src/elm/Page.elm
```diff
@@ -175,19 +169,19 @@ viewMenuFilterDropdown toMsg decoder options =
...
+        [ ]
         buttons
 
 
 viewMenuFilterTabButton : Bool -> (a -> msg) -> Decoder a -> String -> Html msg
 viewMenuFilterTabButton isActive toMsg decoder text_ =
     case isActive of
         True ->
-            button [ class "menu-filter__button menu-filter__button-active", onClick toMsg decoder ]
+            button [ class "menu-filter__button menu-filter__button-active", value text_, onClick toMsg decoder ]
```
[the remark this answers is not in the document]
zacck: Awesome!!!

src/styles/legacy_main.css
zacck: let me know if you need help moving this to tailwind  ...

zacck:
@JerryNyoike  you also need to remeber to bump the `PATCH` version in package.json to register these improvements

zacck:
@JerryNyoike  so while testing your PR I have come across some issues that I would love you to address 

1. The tabs have rounded edges which is good, however the inner edges should not be rounded so that we have the flat line going across see the screengrab below 
2. Clicking on the all offers tabs doesn't work 
3. There is something left over from the old UI if you look at it you will see there is some text saying `filters`

Screenshot 2019-12-18 at 10 07 34

Another recomendation, I would like you to consider implementing  a HTML Lazy strategy for the rendering the list of sales, this is because we want the screen to be lazy when updating the sales since the list could grow to be very huge and I can already notice this when clicking the different tabs 

To learn how to do take a look at this tutorial 
https://guide.elm-lang.org/optimization/lazy.html

If you need guidance through this I am available to help you

PR #113, opened 2019-12-16
Ask button removed. This closes issue #110

### LINKED ISSUE #110: Remove the "Ask" button
After the MUDA pilot tests we realised that we should remove the chat feature for now as mentioned in the [IP 19](https://github.com/cambiatus/improvement-proposals/issues/19)

This button is inside a Shop sale, for example: https://app.cambiatus.io/shop/17

Just remove the button, pretty straightforward

namoscagnm:
Ask menu is removed, and corresponding translations also.
Corresponding Msg, update function and msgToString updated.

closes #110

### DIFF
--- a/src/elm/Page/Shop/Viewer.elm
+++ b/src/elm/Page/Shop/Viewer.elm
@@ -484,13 +484,6 @@ viewCard session card model =
                       else if card.sale.units <= 0 && card.sale.trackStock == True then
                         div [ class "flex -mx-2 md:justify-end" ]
                             [ button
-                                [ class "button button-secondary w-1/5 mx-2"
-                                , onClick (ClickedAsk card.sale)
-                                ]
-                                [ Avatar.view (getIpfsUrl session) card.sale.creator.avatar "h-6 w-6 mr-2"
-                                , text_ "shop.ask"
-                                ]
-                            , button
                                 [ disabled True
                                 , class "button button-disabled w-4/5 mx-2 md:w-2/5"
                                 ]
@@ -509,13 +502,6 @@ viewCard session card model =
                       else
                         div [ class "flex -mx-2 md:justify-end" ]
                             [ button
-                                [ class "button button-secondary w-1/5 mx-2"
-                                , onClick (ClickedAsk card.sale)
-                                ]
-                                [ Avatar.view (getIpfsUrl session) card.sale.creator.avatar "h-6 w-6 mr-2"
-                                , text_ "shop.ask"
-                                ]
-                            , button
                                 [ class "button button-primary w-4/5 mx-2 md:w-2/5"
                                 , onClick (ClickedBuy card.sale)
                                 ]
--- a/src/elm/Page/Shop/Viewer.elm
+++ b/src/elm/Page/Shop/Viewer.elm
@@ -471,7 +471,7 @@ viewCard session card model =
                         div [ class "flex -mx-2 md:justify-end" ]
                             [ button
                                 [ disabled True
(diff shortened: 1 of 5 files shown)


zacck:
@namoscagnm  Thank you so much for taking this one down, could you please edit you PR description with `Closes #110` to connect the PR to the relevant issue, I am running your code and will report back in a minute or so

namoscagnm:
Hi.
Sorry, I do not know how to detect smaller screens, or the code responsible for it. 
- I assume this change will be for every situation, not only small screens?
- Can you point me in the code where to do it? I’m not sure if on the css file, which I don’t understand completely, or the View.elm file, in which case I would have to use a styled button? Do you have a sample from this codebase or somewhere else where I could learn it? Thank you!

lucca65:
@namoscagnm the code responsible for it is the CSS. We have a css file but you don't have to interact with it all, since we use Tailwind. Please take a look at their docs https://tailwindcss.com

on Elm you can use normal CSS classes:

```elm
div [class "asdf"] []
```

That is, the `div` function takes two arguments, both arrays. Function `class` takes one argument, which is a string. You can just add tailwind classes as needed.

You can also look at this file for reference https://github.com/cambiatus/frontend/blob/master/src/elm/Page/Shop/Editor.elm#L438

src/elm/Page/Shop/Viewer.elm
```diff
@@ -484,13 +470,6 @@ viewCard session card model =
...
                             [ button
-                                [ class "button button-secondary w-1/5 mx-2"
-                                , onClick (ClickedAsk card.sale)
-                                ]
-                                [ Avatar.view (getIpfsUrl session) card.sale.creator.avatar "h-6 w-6 mr-2"
-                                , text_ "shop.ask"
-                                ]
-                            , button
                                 [ disabled True
                                 , class "button button-disabled w-4/5 mx-2 md:w-2/5"
```
zacck: Hi @namoscagnm  no worries about showing you so our application uses tailwindcss for styling which you can find out more about here https://tailwindcss.com/docs/margin/#app this page will also show you how to set margins however the easiest way for you to do this would be to add this rule `sm:mx-auto` to this line(#474) and (#491) what this does is it adds an auto margin on smaller devices to know more about using this you would need to look into tailwind but you can complete this without doing that.
Please let me know if I can help further

namoscagnm:
Ok, I updated and tested by making my browser width shorter. It seems to have worked. However, if I make it extremely short, things are not so well aligned anymore. I dont know whether its relevant or not, because I might have shrunk the window in my test more than a real phone would do.

lucca65:
we support a minimum width of 320px, if it works fine with that it is ok

namoscagnm:
1. 

640 pixels width: image

2.

Less than 640 px pixels width:
image

Is more tunning needed?

zacck:
@namoscagnm what you can do is
1. Right click and press inspect 
2. Then from here at the top you will have a drop down like the one below 
Screenshot 2019-12-16 at 18 35 23
3. If the application works on the iphone 6/7 size then it is perfect!!!

namoscagnm:
Thanks @zacck 
I played with the parameters and now the view on iPhone 6 7 8 looks fine. For this I just replaced all styling as you can see, instead of adding.

Please tell me if there are any unexpected side effects. Thanks!

zacck:
@namoscagnm  thank you so much! this is awesome!

namoscagnm:
Thank you too! It was nice for me to participate while I learn ELM, and thank for the support too :)
Lets see how I go in the next ones, hehe!!

PR #116, opened 2019-12-21
This closes issue #103

### LINKED ISSUE #103: Username on card Shop
Muda users missed the name of the person selling a product in the Shop, just the photo was not enough. 

So the improvement is to add the first username on the cards...

**MOBILE**
Captura de Tela 2019-12-09 às 09 56 51

Captura de Tela 2019-12-09 às 10 16 03

**DESK**
Captura de Tela 2019-12-09 às 09 56 37

Captura de Tela 2019-12-09 às 10 16 16

**Interface attributes**

https://www.figma.com/file/idejYyPJDeoqJvodGUCoiV/Cards?node-id=0%3A1

and

https://www.figma.com/file/OMN5j9HUcRQwFbMcVqaBwi/Productpage?node-id=0%3A1

namoscagnm:
Closes #103 

Hi, I added creatorId on the shop overview list and also before the buy button, but you have to check:
- If the CSS is ok
- I couldnt add the user name, because even though the code seems to include it via Sale.avatar on line:
https://github.com/cambiatus/frontend/blob/5580193af833ecc3e30dc5041d53d2446f0b9f8d/src/elm/Shop.elm#L116
and Card seems to contain Sale from line:
https://github.com/cambiatus/frontend/blob/5580193af833ecc3e30dc5041d53d2446f0b9f8d/src/elm/Page/Shop.elm#L141
, I got compile errors when trying to use anything that is member of based on card.sale.creator.avatar. For instance, trying to use
`Eos.nameToString card.sale.creator.avatar.name`
will produce the error: 
**Eos.nameToString card.sale.creator.avatar.name**

From the code i know that Sale has creator.name, and that Card is taking from Sale, but I dont know why this is actually not working.

Thanks,


namoscagnm:
Hi, I have made an update. Information is being show also in smartphone screens and design is a little better, but still some thing missing, as you can see in picture.

image

I need help on following items:
- Setting the vertical spacing as the designer specified
- Finding the full user name. I tried some things as I said before, but I still have no clue on how to do it.

Can you provide some support? Thanks

lucca65:
Hey @namoscagnm maybe you need a little explanation on how to get data using graphql and how Elm handles it.

First of all, please open this URL: https://api.cambiatus.io/api/graphiql

Fill in this query:

```graphql
query {
  sales(input: { account: "lucca" }) {
    description
    price
    community {
      name
    }
    creator {
      name
      account
    }
  }
}
```

There is this section with a `creator`, nested to it we have two fields: `account` and `name`. This means that you are asking for the GraphQL API to get those two informations for you. 

Now, we use a library that uses the ability to introspect types that GraphQL offer us to build Elm types that correspond to our GraphQL API. That is, we now have the compiler to help us to only use fields that exists on our API, so we don't have typos or JSON errors that are super common during JS project development. All those generated types resides inside the `/Bespiral` folder. (trivia: BeSpiral was our old name before Cambiatus and we still havent migrated the naming on this folder). There we have several domain types.

Now. As you noticed we currently are only asking for the avatar on our `SelectionSet`. SelectionSets are just normal queries, that correspond to the one I've sent above. So you will have to modify this function to ask for more info to GraphQL.
Screen Shot 2019-12-23 at 15 37 10

Also take a look at this line: https://github.com/cambiatus/frontend/blob/5580193af833ecc3e30dc5041d53d2446f0b9f8d/src/elm/Shop.elm#L103

You see that the first argument this function receives is a `Sale`. This type is defined here: https://github.com/cambiatus/frontend/blob/5580193af833ecc3e30dc5041d53d2446f0b9f8d/src/elm/Shop.elm#L23

This type and the `|> with . . . ` chain function calls are identical, each `with` for a property in this map. You'll have to modify this type to have the information your screen will need, and make the `SelectionSet` fill it in for you. This way you'll have the information you need. 

Try to keep in mind that Elm is a very simple language, the looks can be deceiving. Elm has no elements common in prototype languages such as JS nor inheritance as in Java, so Types contain what they say they contain, nothing else (in C# and Java it is very common to have types extending each other, sometime by a external lib such as Entity Framework or Hibernate). So when you try to call `Eos.nameToString card.sale.creator.avatar.name` it will not work as avatar is a not a Map type.

Hope it helps, let me know if you need further instructions

zacck:
@namoscagnm could we assist you further here?

lucca65:
I'll finish this in another PR, since @namoscagnm won't be able to do it in the next 1 or 2 weeks

PR #124, opened 2020-01-14
[WIP] - Implement github workflow for the project

zacck:
## Changes Proposed 

- [x] Implement github workflow for the project 
- [x] Pubish slack webhook upon workflow completion


zacck:
That took a minute 👀

zacck:
Hi @lucca65  I think I have CI technicals closely figured out but I think this now makes me open an issue in Improvement proposals I would like us to describe our workflow so that we can have one standard one through out the organization

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.7.2",
   "private": true,
   "engines": {
-    "node": "11.15.0"
+    "node": "12.0.0"
```
lucca65: @zacck have you tried to build our docker image after this change? we are fixed in `11.15.0` mainly because the `docker build` cmd fails with other node versions

.github/workflows/on-push-action.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Get deps
+        run: npm install
+
+      - name: Build asset
+        run: npm build
+
+      - name: Publish Slack Notification
+        uses: rtCamp/action-slack-notify@master
+        env:
+          SLACK_WEBHOOK: ${{ secrets.FRONT_END_WEBHOOK }}
```
lucca65: seems simpler than circleci but we still not building the required docker image

package.json
zacck: I checked the builds but not in docker ... let me see how it goes

zacck:
@lucca65  docker builds successfully, also modified to use an updated image

lucca65:
@zacck is it possible for us to build the docker image and upload it to docker hub?

zacck:
@lucca65  yes ill add a directive to run on merge to master

zacck:
@lucca65  I added some more more directives to ship this.... let me know what you think

lucca65:
sweet @zacck!

PR #183, opened 2020-03-02
Validate sale price field to only include numbers

### LINKED ISSUE #137: Input price shouldn't accept letters, only numbers
**Describe the bug**
Input price is accepting letters

**To Reproduce**
1. Go to Shop
2. Click on 'Create an offer
3. Go to Price input
4. Type letters
5. See the error

**Expected behavior**
Only accept numbers

**Screenshots**
inputNumber

app.cambiatus.io

RodrigoLeiteF:
Closes #137 

Before updating the model we check every character in the sale price and filter out the non-digit ones.

### DIFF
--- a/src/elm/Page/Shop/Editor.elm
+++ b/src/elm/Page/Shop/Editor.elm
@@ -712,7 +712,7 @@ update msg model loggedIn =
             model
                 |> updateForm
                     (\form ->
-                        { form | price = updateInput value form.price }
+                        { form | price = updateInput (getNumericValues value) form.price }
                     )
                 |> UR.init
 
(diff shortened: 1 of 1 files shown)


src/elm/Page/Shop/Editor.elm
```diff
@@ -1148,6 +1148,15 @@ encodeUpdateForm sale form =
         ]
 
 
+getNumericValues : String -> String
+getNumericValues value =
+    value
+        |> String.toList
+        |> List.filter (\x -> Char.isDigit x)
```
lucca65: No need to send a function, you can curry it:

```
|> List.filter Char.isDigit
```

RodrigoLeiteF: Good catch. Fixed

PR #191, opened 2020-03-11
Add friendlier message when Javascript is disabled

### LINKED ISSUE #90: Display a user friendlier message when Javascript is disabled
I always ashamed of the user message I run into when my  browser has Javascript disabled and I am trying to access the website. 
All one sees is a single line of text at the top right of the screen saying one needs to enable Javascript 

Since we are going to be catering for low powered device usage I am fairly sure this will occur/ has occured in the field a number of time and Ideally the way we can mitigate this is by showing  some styled HTML content that identifies our application and perhaps helps the user to correct the settings issue 

**Screen In Question** 
Screenshot 2019-11-26 at 09 42 17

RodrigoLeiteF:
# What issue does this PR close
Closes #90 

Adds friendlier message when javascript is disabled, as per https://www.figma.com/file/9BX6vISFLh3Ns0HtVx2XJX/Special-pages?node-id=22%3A1059

1. Go to `about:config`
2. Toggle `javascript.enabled` to `false`
3. Visit the home page


zacck:
Hey @RodrigoLeiteF  could you please look at the sizing of the SVG at the bottom of the page? Currently it loads at a size that squeezes the top content out of the page  see my screenshot below 

Screenshot 2020-03-12 at 08 50 20

Ideally we should have it look like the following, which I achieved by reducing the height of the SVG, you may need to confer with @heltonlr  on how much it should be scaled down by 

Screenshot 2020-03-12 at 08 52 49

RodrigoLeiteF:
@zacck That's odd, I can't reproduce the issue you described (in 1920x1080 resolution). Which browser and version did you use?

public/noscript.css
```diff
@@ -0,0 +1,113 @@
+
+       @font-face {
```
lucca65: please remove the initial tabs from all the lines

public/index.html
```diff
@@ -35,7 +35,28 @@
 
   <body class="flex flex-col h-full">
     <noscript>
-      <h3>You need to enable JavaScript to run this app.</h3>
+      <link href="noscript.css" rel="stylesheet">
```
lucca65: in this case we will always be loading noscript.css anyway. at this point we should have full access to tailwind, so why not use it

RodrigoLeiteF: I don't think we do. When I access the page with javascript disabled tailwind classes don't seem to work. According to Firefox Tailwind classes seem to be inlined somewhere, which would explain this behavior. With `noscript.css` being loaded under the noscript tag, it will only be downloaded and executed in that block.

image

lucca65: I see @RodrigoLeiteF!

RodrigoLeiteF:
@lucca65 I made some changes, could you please test again? Everything seems to be correct on Firefox and Chromium for me. Unfortunately there aren't any good Webkit browsers for Linux afaik

lucca65:
Will do @RodrigoLeiteF, but you can use [Midori](https://www.midori-browser.org/about/) to test webkit

public/noscript.css
```diff
@@ -0,0 +1,121 @@
...
+    padding: 0;
+}
+
+noscript {
+    font-family: Gotham Rounded;
+    font-style: normal;
+    font-size: 14px;
+    display: flex;
+    justify-content: center;
+    background: #E5E5E5;
```
lucca65: @RodrigoLeiteF just quick checked Figma and the color for the background is #F8F8F8 (https://www.figma.com/file/9BX6vISFLh3Ns0HtVx2XJX/Special-pages?node-id=22%3A1059)

PR #197, opened 2020-03-16
Fix container boundaries on view transfer screen

lucca65:
Fixes container size on transfer screen

Screen Shot 2020-03-16 at 19 40 44
Screen Shot 2020-03-16 at 19 40 41

### DIFF
--- a/src/elm/Page/Transfer.elm
+++ b/src/elm/Page/Transfer.elm
@@ -97,14 +97,14 @@ view loggedIn model =
 
         InvalidId invalidId ->
             div [ class "container mx-auto px-4" ]
-                [ Lazy.lazy viewHeader loggedIn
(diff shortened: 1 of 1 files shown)


lucca65:
@Pancakem I was unable to test this PR. Looks like during the merge it was not done correctly and the conflicts were solved by choosing the wrong versions (looking by the errors generated)

Looks like you had to add pickup from a very old version. My advice would be to reimplement since cherry picking git commits can be hard. 

Although, if you want do dive deep into git, merge types, and cherry picking I recommend the following blog: https://nvie.com/posts/a-successful-git-branching-model/

its old but goes into detail on a git model, and later discuss how to handle history changes like we are having now. 

Either way its up to you to decide!

PR #199, opened 2020-03-17
Enhancement/keyboard enter fix

### LINKED ISSUE #100: Forms should works with enter (keyboard)
**Describe the bug**
Forms do not send the data when user click enter on the keyboard.

**To Reproduce**
1. Go to any form
2. Enter the data
3. Click enter on the keyboard

**Expected behavior**
- When user click enter on the keyboard , send the data.

**Screenshots**
--

- muda.cambiatus.io

 - not apply

 - Device: Samsung (several) and Xiaomi;
 - OS: Android;
 - Browser: Chrome;
 - Version: --

**Additional context**
--

Pancakem:
Closes #100 

Add enter key handling to submit forms.

1. Go to any form
2. Enter the data
3. Click enter on the keyboard


Pancakem:
@lucca65 I have an issue with adding date here https://github.com/Pancakem/frontend/blob/enhancement/keyboard-enter-fix/src/elm/Page/Community/ActionEditor.elm#L909. Any suggestions on how I can do it?

lucca65:
instead of calling save action, call ValidateDeadline

Pancakem:
@lucca65 I have made the changes

src/elm/Page/Dashboard/Community.elm
```diff
@@ -711,6 +728,18 @@ update loggedIn msg model =
             UR.init { model | autoCompleteState = updated }
                 |> UR.addCmd cmd
 
+        PressedEnter val ->
+            case val of
+                True ->
+                    UR.init model
+                        |> UR.addCmd
+                            (Task.succeed ClickedTransfer
+                                |> Task.perform identity
```
lucca65: this snippet still uses `case` instead of `if`. 

You'll need to test it, but looks like you should call `ClickedSendTransfer` instead of `ClickedTransfer`

src/elm/Page/Profile.elm
```diff
@@ -830,6 +842,18 @@ update msg model loggedIn =
                             [ ( "name", Encode.string "printAuthPdf" ) ]
                     }
 
+        PressedEnter val ->
+            case val of
```
lucca65: still using case

src/elm/Page/Dashboard/Community.elm
Pancakem: Hey @lucca65 I have found out why it brings up the submit pin form. 
When `ClickedSendTransfer` is called the second case, https://github.com/Pancakem/frontend/blob/enhancement/keyboard-enter-fix/src/elm/Page/Dashboard/Community.elm#L617, evaluates.

Pancakem: @lucca65 I haven't found a solution for it yet. The pin is still prompted. Can it be an auth issue? When I input the pin when prompted the transfer s completed.

lucca65: @Pancakem taking a deeper look into this and I noticed that it only happens when we press enter to choose which user we want to transfer to. If we use the mouse instead of pressing enter it don't show this behaviour.

This means that we need to make Auth.elm ignore enter until it is visible.

The way would be to figure out how the Auth module knows if it can show the auth modal or no, and use that same method to register the `PressedEnter` event (I'm talking the one that exists inside `Auth.elm` not the one you implemented here)

lucca65: @Pancakem I've just committed the solution into your branch. After trying to do this I realize that it required more knowledge on the login process, maybe we can take this slower with you, there are plenty of other issues that might require that knowledge too, so we have plenty of opportunities!

PR #212, opened 2020-03-31
Unify feedback messages

### LINKED ISSUE #196: Refactor: unify feedback messages
Although in our designs we have a single way to pass success failure feedback, on the implementation we don't. We should unify feedback messages

Add a new key in the shared map with messages, allowing us to configure if it is success or failure. Use our default designs: https://www.figma.com/file/oZJUIyHRmLyLIKDxYebvLg/Form-elements?node-id=7%3A296

RodrigoLeiteF:
Closes #196 

Adds a centralized API for giving users feedback

TODO

### DIFF
--- a/src/elm/Auth.elm
+++ b/src/elm/Auth.elm
@@ -4,6 +4,7 @@ import Api
 import Asset.Icon as Icon
 import Browser.Dom as Dom
 import Eos.Account as Eos
+import Feedback
 import Flags
 import Html exposing (..)
 import Html.Attributes exposing (..)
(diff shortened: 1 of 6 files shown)


src/elm/Auth.elm
```diff
@@ -146,87 +147,107 @@ maybePrivateKey model =
 
 -- VIEW
 
+{-| You're gonna want a comment explaining how this works I reckon. -}
+type alias Handlers msg =
```
lucca65: Yeah, looks like you are changing the way we map Msgs. Maybe we can talk in more depth about this on a video call

src/elm/Feedback.elm
```diff
@@ -0,0 +1,32 @@
...
+    , success : Bool
+    }
+        
+
+-- VIEW
+
+
+getBackgroundColor : Bool -> String
+getBackgroundColor isSuccess =
+    if isSuccess == True then
```
lucca65: it is redundant. `if isSuccess then` is equivalent

src/elm/Main.elm
```diff
@@ -197,6 +198,7 @@ type Msg
     | GotUpdatedBalances (Result Http.Error (List Community.Balance))
     | GotShopViewerMsg ShopViewer.Msg
     | GotTransferScreenMsg Transfer.Msg
+    | ShowFeedback Feedback.Model
```
lucca65: Why not just make it a `GotPageMsg` message since we already have this generalised structure?

src/elm/Page/Dashboard.elm
```diff
@@ -122,15 +129,19 @@ view loggedIn model =
...
                     , div [ class "text-indigo-500 ml-2 font-medium" ]
                         [ text (profile.userName |> Maybe.withDefault (Eos.nameToString profile.account))
                         ]
                     ]
-                , viewBalances loggedIn communities
-                , viewVerifications loggedIn.shared model
-                , viewSections loggedIn model
+                , viewBalances loggedIn communities |> Html.map h.tagger
+                , viewVerifications loggedIn.shared model |> Html.map h.tagger
+                , viewSections loggedIn model |> Html.map h.tagger
```
lucca65: Hmm mapping the Dashboard messages seems very weird, it should be the other way around

lucca65: I mean, local messages shouldn't need any mapping, outside messages might need. Also the feedback should be global, that is, something outside of the page itself. Like the menu, for example. Its global and don't live in the individual page

src/elm/Page/Login.elm
```diff
@@ -92,6 +95,12 @@ update msg model guest =
             shared.language
     in
     case msg of
+        ShowFeedback _ ->
+            let
+                _ = Debug.log "I shouldn't be called" ""
```
lucca65: lol

src/elm/Page/Login.elm
```diff
@@ -132,3 +141,5 @@ msgToString msg =
     case msg of
         GotAuthMsg subMsg ->
             "GotAuthMsg" :: Auth.msgToString subMsg
+
+        _ -> []
```
lucca65: In order to avoid stuff like this, please configure your text editor to add a newline at the end of the file when it saves, its a feature every editor should have an option by default (without extensions)

src/elm/Session/LoggedIn.elm
```diff
@@ -125,6 +126,7 @@ type alias Model =
     , showAuthModal : Bool
     , auth : Auth.Model
     , balances : List Balance
+    , feedback : Maybe Feedback.Model
```
lucca65: I like this, but `Maybe` isn't good. We can achieve the same result but with lot of more semantics by introducing a type:

```elm
type Feedback = 
     Dismissed
     | Enabled Feedback.Model
```

Same result but now can do cases without having to handle `maybe`. Not to say that it isn't useful, but it will make the debugger easier to read in the future since we will have some error like "you didn't implement all possible values for `Feedback`" instead of "the function is expecting Maybe Feedback.Model, but the provided value is Feedback.Model"

It will also avoid putting `Just` everywhere too

lucca65: If you want to go deeper into types you may take a look at Theorem proving languages like Coq or Idris. Their compiler and type notion is much improved.

I also recommend this exercise in Haskell, that is not specific to the discussion on `Maybe vs types` but does highlight how we can use the compiler to help us keep context in our application: 

https://reasonablypolymorphic.com/blog/typeholes/index.html

RodrigoLeiteF:
Closing this in favor of #218

PR #214, opened 2020-04-02
Add validation for manual verification

### LINKED ISSUE #85: Require at least two verifiers in the verifier list
Make sure that if we have a list of verifiers active on the action, it is required of it to have at least two users.

amiskov:
Closes #85 

- add validation function for the list length (>= a given value);
- wrap `List Profile` (verifiers) in `Validation` type and check that length of this list is >= than min. number of votes.

1. Create new or edit existing Action.
2. Select "Manual" verification.
3. Press "Create"/"Edit" button with no verifiers selected of with number of verifiers less than min. number of votes.
4. There should be an error like this:
    image

Also works for other min. numbers of votes:
image

### DIFF
--- a/public/translations/en-US.json
+++ b/public/translations/en-US.json
@@ -47,6 +47,7 @@
         "negative": "The number must be negative",
         "greater_than": "The number must be greater than {{base}}",
         "greater_than_or_equal": "The number must be greater than or equal to {{base}}",
+        "length_greater_than_or_equal": "Choose at least {{count}} items",
         "lower_than": "The number must be lower than {{base}}",
         "lower_than_or_equal": "The number must be lower than or equal to {{base}}"
       },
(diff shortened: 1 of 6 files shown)


public/translations/en-US.json
```diff
@@ -323,6 +324,7 @@
         "validation_on_detail": "the action will be available until a date, quantity or both",
         "verification_label": "Does it need verification?",
         "verifiers_label": "Choose the verifiers",
+        "verifiers_label_count": "Choose at least {{count}} verifiers",
```
lucca65, line 327: We are still missing translations for other languages. Usually we just use google translate, we had little errors with it. Currently we support `en-US`, `pt-BR`, `es-ES` and `cat-CAT`

here is the Portuguese one: `Escolha ao menos {{count}} verificadores`

public/translations/en-US.json
```diff
@@ -47,6 +47,7 @@
         "negative": "The number must be negative",
         "greater_than": "The number must be greater than {{base}}",
         "greater_than_or_equal": "The number must be greater than or equal to {{base}}",
+        "length_greater_than_or_equal": "Choose at least {{count}} items",
```
lucca65, line 50: We are still missing translations for other languages. Usually we just use google translate, we had little errors with it. Currently we support `en-US`, `pt-BR`, `es-ES` and `cat-CAT`

here is the Portuguese one: `Escolha ao menos {{count}} items`

amiskov:
Added missing translations and squashed commits into one.

lucca65:
Merged! Thanks for the contribution @amiskov

PR #218, opened 2020-04-06
Centralized API for Feedback mesasges

### LINKED ISSUE #196: Refactor: unify feedback messages

RodrigoLeiteF:
Closes #196


src/elm/Session/LoggedIn.elm
```diff
@@ -204,6 +218,24 @@ view thisMsg page ({ shared } as model) content =
...
+        , style "background-color" (getBackgroundColor feedback.success)
+        ]
+        [ span [ class "ml-auto invisible" ] []
+        , span [ class "flex items-center text-sm h-10 leading-snug text-white font-bold" ]
+            [ text feedback.message ]
+        , span
+            [ class "ml-auto text-lg text-white font-bold mr-5 cursor-pointer"
+            , onClick HideFeedback
+            ]
+            [ text "X"
```
lucca65, line 234: Its best we use the `x` icon so we are consistent with the rest of the application.

We can have

```elm
Icons.close ""
```

src/elm/Main.elm
```diff
@@ -488,6 +488,16 @@ updateLoggedInUResult toStatus toMsg model uResult =
 
                         _ ->
                             ( m, cmds_ )
+
+                LoggedIn.ShowFeedback x ->
```
lucca65, line 492: we should have a better name for this value

src/elm/Session/LoggedIn.elm
```diff
@@ -478,6 +518,15 @@ viewFooter _ =
         ]
 
 
+getBackgroundColor : Bool -> String
+getBackgroundColor isSuccess =
+    if isSuccess == True then
+        "#8ACC9E"
+
+    else
+        "#DB1B1B"
```
lucca65, line 527: we should use tailwind classes for that `green` and `red`. Actually maybe we don't need a function to encapsulate this `if`

RodrigoLeiteF: I tried those classes but for some reason they wouldn't work. Maybe we're using an outdated version?

PR #222, opened 2020-04-14
Refactor feedback messages to comply with new system

RodrigoLeiteF:
Adds positive feedback for the following actions
1. Editing a profile
2. Creating and editing an objective
3. Creating and editing an action
4. Creating, editing and deleting an order

Open the relevant pages and complete the described actions. A feedback
should appear at the top.


lucca65:
Also what are your thoughts on changing the type for Show message?

instead of using a record  with `ShowFeedback { message : String, success : Bool }` why not `ShowFeedback String Bool`? I think its simple enough to justify the change. 

What do you think @RodrigoLeiteF?

I would love to know @amiskov thoughts on this too

RodrigoLeiteF:
> Also what are your thoughts on changing the type for Show message?
> 
> instead of using a record with `ShowFeedback { message : String, success : Bool }` why not `ShowFeedback String Bool`? I think its simple enough to justify the change.
> 
> What do you think @RodrigoLeiteF?
> 
> I would love to know @amiskov thoughts on this too

I personally prefer using a record with named parameters. Ime it's almost always hard to infer what an unnamed boolean does. For example, reading `ShowFeedback "Action created successfully" True` or `ShowFeedback "Internal server error" False` isn't very intuitive I think.

amiskov:
> instead of using a record with `ShowFeedback { message : String, success : Bool }` why not `ShowFeedback String Bool`?

Looks like a simple functions to me: `showSuccess "Successfully done!"`, `showFailure "Error..."`. But if it has to be a type, then I agree with @RodrigoLeiteF: explicit is better than implicit :)

RodrigoLeiteF:
After some discussion with friends I think it would be best to use a type for `Success | Failure` instead, which would result in `ShowFeedback "Successfully claimed action" Success`. We use a similar pattern for the feedback status on LoggedIn.elm (custom type as opposed to Maybe) :)

lucca65:
+1 in the `ShowFeedback Status String` approach (maybe its best to leave the message as the second param)

lucca65:
Looks good @RodrigoLeiteF, you'll need to deploy it so Helton can test it out

Issue #211, opened 2020-03-31
Error while trying to export the private key


lucca65:
**Describe the bug**
App don't respond when trying to export the private key/PDF from the profile.

In development mode we get an JS error, displayed in the print screen below

**To Reproduce**
1. Go to 'Profile'
2. Click on the key icon
3. See error

**Expected behavior**
A PDF download of the private key

**Screenshots**
<img width="1057" alt="Screen Shot 2020-03-30 at 22 20 37" src="https://user-images.githubusercontent.com/1082127/77976937-b1369900-72d4-11ea-8fcc-6acf3092178b.png">

**Cambiatus Env you are using**
app.cambiatus.io

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A

**Additional context**
Discovered by @zequez

amiskov:
As far as I can see, the problem is that there are 2 buttons in [Page/Profile.elm#L215](https://github.com/cambiatus/frontend/blob/master/src/elm/Page/Profile.elm#L215), one above the other. In browser it renders like there's only 1 button. Here I shifted one of them in browser tools to demonstrate:

![image](https://user-images.githubusercontent.com/140053/78333068-d60b6480-7591-11ea-9521-f7ac9e4c815c.png)

The button above (which we actually click) tries to download PDF and fails, because PDF generator tries to find the text in HTML elements, which presented only on "Congratulations" page (page appears after the registration of the user).

The button below works fine, but it only shows the private key without generating PDF.

It's unclear, what is the right way to handle click on this button. Should we download PDF or just show the private key?

lucca65:
nice investigation @amiskov. Yeah, profile screen is one that we still have to rework. We started a redesign that is still ongoing for some less used screens. Profile page is one of them. The correct behaviour would be to download the PDF.

I think the right way to do it is, instead of copying from the existing elements, we'll need to change that port to take the private key from the local storage, and generate the PDF from it. Then it will work on both scenarios.

amiskov:
I started working on this issue. Need to do some exploration at first and then I'll create a WIP PR and likely ask some questions.

amiskov:
I think I need Profile page design mock-ups to handle this issue properly. @lucca65, can you please provide me these mock-ups in Figma?

amiskov:
> I think the right way to do it is, instead of copying from the existing elements, we'll need to change that port to take the private key from the local storage, and generate the PDF from it. Then it will work on both scenarios.

[...]

PR #224, opened 2020-04-20
Improve Login and Registration UI/UX

### LINKED ISSUE #151: New sign In/create account interface and usability improvement
With the new **sign In/create account** interface and **usability improvement** we will probably correct some difficulties that our pilots had when creating their accounts.

**All flows and interface rules on Figma**

_Create account:_
https://www.figma.com/file/vqHVOvugUksHKlhJvNfaFq/LoginCreateAccount-desk?node-id=1%3A259

_Sign In:_
https://www.figma.com/file/vqHVOvugUksHKlhJvNfaFq/LoginCreateAccount-desk?node-id=1%3A13100

_Default erros elements:_
https://www.figma.com/file/vqHVOvugUksHKlhJvNfaFq/LoginCreateAccount-desk?node-id=1%3A4539

_Responsive rules_
https://www.figma.com/file/vqHVOvugUksHKlhJvNfaFq/LoginCreateAccount-desk?node-id=82%3A79

amiskov:
Closes #151 

- Split login in 2 steps: passphrase + pin.
- Implement [new UI changes from Figma](https://www.figma.com/file/47BEsGEMSWfhTm3zC0ugyA/Main-flow?node-id=0%3A1).
- Code refactoring according new requirements.

Pages for Login and Registration should have new, improved UI.


src/elm/Auth.elm
```diff
@@ -31,13 +31,24 @@ import Json.Encode as Encode exposing (Value)
...
 import UpdateResult as UR
 import Utils
-import Validate exposing (Validator, fromValid, ifNotInt, validate)
+import Validate exposing (Validator, validate)
+
+
+{-| Authentication of a user with Passphrase or Private Key.
+
+  - Passphrase consists of 12 unique words given to the user during the registration. Only users knows this phrase.
+  - Private Key (PK) is a hashed analogue of a Passphrase, we identify the user by the PK in the backend.
```
lucca65, line 45: actually we identify the user by the Public Key, not the Private Key!

src/elm/Auth.elm
```diff
@@ -176,21 +179,45 @@ passphraseValidator =
...
+validatePinNumber pin =
+    let
+        hasCorrectLength =
+            String.length pin == 6
+
+        hasOnlyDigits =
+            String.all Char.isDigit pin
+    in
+    if not hasCorrectLength || not hasOnlyDigits then
+        [ InvalidEntry PinConfirmation "PIN must be 6 digits (TR)" ]
```
lucca65, line 192: I know this is just a WIP, but please make sure the errors are localised! 😉

src/elm/Auth.elm
amiskov: Oh, I see 😅 Thank you.

So, here is the right flow: 12-words → private key → public key. And then we use the last one for communication with the backend. Right?

@lucca65 How do you think, was it a good idea to add this description? I think it may be useful for newcomers and for future us. As for me, I spent quite a time understanding the code and this kind of documentation could help me a lot.

src/elm/Auth.elm
amiskov: Sure! Those `(TR)` in texts are reminders for me to add translations.

src/elm/Auth.elm
lucca65: That's exactly right @amiskov! 💪 In the end the private key never go to the internet in any way or form. 

About the description: +1000 I love this approach of documenting as we go along, feels much more natural and helpful!

src/elm/Auth.elm
lucca65: nice! keep the good work! 💪

src/elm/Page/Register.elm
```diff
@@ -269,52 +271,64 @@ viewServerErrors problems =
...
-    , validation : ValidatedField
+    , currentValue : String
+    , fieldName : ValidatedField
     }
 
 
-viewField : Shared -> Field -> (String -> FormInputMsg) -> List (Attribute FormInputMsg) -> List Problem -> Html Msg
-viewField ({ translations } as shared) { translationSuffix, isDisabled, id_, validation } msg extraAttrs problems =
+viewField : Shared -> Model -> Field -> (String -> FormInputMsg) -> List (Attribute FormInputMsg) -> Html Msg
+viewField ({ translations } as shared) { form, problems } { translationSuffix, isDisabled, currentValue, fieldName } msg extraAttrs =
```
amiskov, line 280: `form` is redundant here.

src/elm/Page/Register.elm
```diff
@@ -193,43 +193,45 @@ view guest model =
...
                 , viewServerErrors model.problems
-                , viewField shared
+                , viewField
+                    shared
+                    model
                     (Field
                         "register.form.name"
                         isDisabled
-                        "name"
+                        model.form.username
```
amiskov, line 207: It's convenient to use translation suffix as an `id` in HTML. Here I pass current value of the field to use it in view rendering.

src/elm/Session/Guest.elm
```diff
@@ -71,9 +71,13 @@ subscriptions model =
...
 
 
+{-| Page types the Guest can access.
+-}
 type Page
-    = Other
+    = Register
+    | Login
     | Shop
+    | Other
```
amiskov, line 80: I added Register and Login here to distinguish some styles for those pages for guest.

amiskov:
I think it's ready for review. @lucca65  @RodrigoLeiteF please, take a look!

src/elm/Session/Guest.elm
```diff
@@ -86,101 +90,122 @@ view thisMsg page ({ shared } as model) content =
...
+                    , style "background-color" "#EFF9FB"
+                    , style "background-position" "center bottom"
+                    , style "background-size" "auto 80%"
+                    , style "background-image" "url(images/auth_bg_full.png)"
+                    ]
+                    [ viewQuote
+                        { photoSrc = "images/woman.png"
+                        , name = "Victoria Lane"
+                        , occupation = "Founder"
+                        , quote = "We are designing for a global user base, we can get feedback not just from drivers we see in San Francisco"
```
lucca65, line 112: We should use an actual quote here, from an actual user, using an actual photo lol

@heltonlr can you help us on that? It also need to be localised

src/elm/Session/Guest.elm
```diff
@@ -86,101 +90,122 @@ view thisMsg page ({ shared } as model) content =
...
+            ]
+        ]
+
+
+type alias Quote =
+    { photoSrc : String
+    , name : String
+    , occupation : String
+    , quote : String
+    }
```
lucca65, line 185: this is very smart @amiskov

src/styles/legacy_main.css
```diff
@@ -3706,9 +3532,6 @@ textarea.form-input {
         margin-left: -16px;
     }
 
-    .card__auth__register__prompt {
-        width: unset;
-    }
```
lucca65, line 3711: The happiness!! Soon we will be able to remove this file completely

amiskov:
> There is no style for the error message on the account input. I've checked @heltonlr's designs and we really don't have it on there, but there is a pattern for error messages that we follow, you can see it here: https://www.figma.com/file/oZJUIyHRmLyLIKDxYebvLg/Form-elements?node-id=1%3A908

I actually wrote styles for those errors according the designs from the link. E.g. it works fine [here](http://cambiatus.miskov.ru/register) for me. 

@lucca65 How have you found this problem?  It looks like some CSS wasn't loaded to the browser...

lucca65:
Weird must be something cached! tested on my phone and it worked

src/elm/Page/Profile.elm
```diff
@@ -231,7 +231,9 @@ view_ loggedIn profile model =
                                 , ( "circle-background", True )
                                 , ( "circle-background--primary", True )
                                 ]
-                            , onClick DownloadPdf
+
+                            -- TODO: Should be passphrase, not privateKey
```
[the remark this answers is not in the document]
amiskov: The thing is that the user while downloading PDF from the Profile may get the private key instead of 12 words if the page was reloaded.

This TODO is actually related to the #211 issue (registration and login work fine). I think it's better to handle it there.

src/elm/Session/Guest.elm
amiskov: I've hidden the user quote until we have some actual user quotes. I will add them as a part of #238.

src/elm/Auth.elm
```diff
@@ -189,81 +275,161 @@ view isModal shared model =
...
+                    ]
+                    []
+                , div [ class "input-label pr-1 text-white font-bold mt-1 text-right" ]
+                    [ text <|
+                        let
+                            passphraseWordsCount =
+                                String.fromInt <| List.length (List.filter (not << String.isEmpty) <| String.words model.form.passphrase)
+
+                            _ =
+                                Debug.log "passphraseWordsCount"
```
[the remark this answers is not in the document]
amiskov: Oops... For dealing with this kind of stupid mistakes we can adopt [elm-reivew](https://github.com/jfmengels/elm-review), it may save us some time. I tried this package today and it worked well. I'll create an issue after finishing the current task with some details.

amiskov:
I'm getting used to this well-known refactoring with Elm. To be honest I just have to push myself to stop doing it :) @lucca65 can you please check the code once again?

lucca65:
Glad to read that @amiskov, I really enjoy how confident we are on changing things up with Elm, no way I get the same feeling even with Elixir! I'll check the changes!

src/styles/main.css
```diff
@@ -241,6 +249,37 @@
...
+.lang-menu-open .lang-flag {
+  @apply mr-2;
+}
+
+.min-h-stretch {
+  /* Fill all available vertical space. It works predictable for mobile,
+  because it doesn't take the address bar into account. */
+  @apply min-h-screen;
+  min-height: stretch;
+}
```
lucca65, line 268: I see that you've added new classes for this footer behaviour. We already have some defined that we use in the existing application footer. can't we just use what already exists? maybe there is something different that I'm not aware of

src/elm/Page/Register.elm
```diff
@@ -43,7 +40,7 @@ init guest =
 
 subscriptions : Model -> Sub Msg
 subscriptions _ =
-    Sub.map PressedEnter (Browser.Events.onKeyDown decodeEnterKeyDown)
+    Sub.map KeyPressed (Browser.Events.onKeyDown decodeEnterKeyDown)
```
lucca65, line 43: I like this, `PressedEnter` isn't a good name indeed

src/elm/Page/Register.elm
```diff
@@ -55,6 +52,8 @@ type alias Model =
     , isLoading : Bool
     , isCheckingAccount : Bool
     , form : Form
+    , isAgreedToSavePassphrase : Bool
```
lucca65, line 55: maybe `hasAgreed..`?

src/elm/Page/Register.elm
```diff
@@ -172,108 +151,197 @@ view guest model =
...
-                [ onClick DownloadPdf
-                , class "btn btn__register"
+                [ div [ class "sf-content" ]
+                    [ viewTitleForStep 2
+                    , p
+                        [ class "text-xl mb-3" ]
+                        [ text_ "register.account_created.greet"
+                        , text " "
+                        , strong [] [ text name ]
+                        , text ", "
```
lucca65, line 198: what do you think about adding the spaces and commas to the translated text so we don't have to have those spacing stuff here?

src/elm/Page/Register.elm
```diff
@@ -172,108 +151,197 @@ view guest model =
...
+                        , text_ "register.account_created.last_step"
+                        ]
+                    , p [ class "mb-3" ]
+                        [ text_ "register.account_created.instructions"
+                        ]
+                    , div
+                        [ class "w-1/4 m-auto relative"
+                        , style "left" "1rem"
```
lucca65, line 206: maybe we can achieve the same result if we only have a `left-0` class with a `pl-4` or `ml-4`?

I would prefer to avoid `style` most of the time

src/elm/Page/Register.elm
```diff
@@ -172,108 +151,197 @@ view guest model =
...
+                        , style "max-width" "10rem"
```
lucca65, line 207: here we can use tailwind https://tailwindcss.com/docs/max-width/#app with a new configuration for `xxs`

src/elm/Page/Register.elm
```diff
@@ -283,149 +351,87 @@ viewServerErrors problems =
...
+        ul [ class "bg-red border-lg rounded p-4 mt-2 text-white" ] errorList
 
 
 type alias Field =
     { translationSuffix : String
     , isDisabled : Bool
-    , id_ : String
-    , validation : ValidatedField
+    , currentValue : String
+    , fieldName : ValidatedField
```
lucca65, line 371: maybe just `name` as we already are in the `Field` type?

src/elm/Page/Register.elm
```diff
@@ -519,67 +480,86 @@ type alias UpdateResult =
...
+                                (InvalidEntry Account (tr "error.tooLong" [ ( "maxLength", "12" ) ]))
+                            , ifFalse
+                                (\f -> String.all isValidAlphaNum f.account)
+                                (InvalidEntry Account (t "register.form.accountCharError"))
+                            ]
+                        ]
             in
-            case allProbs of
-                [] ->
+            case validate formValidator form of
```
lucca65, line 557: much better! 👍

src/elm/Auth.elm
```diff
@@ -17,40 +17,36 @@ module Auth exposing
...
+{-| Authentication of a user with Passphrase or Private Key.
+
+  - Passphrase consists of 12 unique words given to the user during the registration. Only users knows this phrase.
+  - Private Key (PK) is a hashed analogue of a Passphrase.
+  - PIN is used to encrypt the Passphrase/PK in the browser. Each time the user logs-in the new PIN is created.
+
+User may use Passphrase and PK interchangeable for logging in, although we push the user forward to use the Passphrase
+because it's more convenient for humans.
+
+-}
```
lucca65, line 49: 👏

src/elm/Auth.elm
```diff
@@ -88,51 +84,140 @@ subscriptions _ =
...
+    | LoggingInWithPrivateKeyAccounts (List Eos.Name) LoginFormData
     | LoggedInWithPrivateKey PrivateKey
     | LoginWithPin
     | LoggingInWithPin
     | LoggedInWithPin PrivateKey
 
 
-type alias PrivateKeyLogin =
-    { privateKey : String
+type alias LoginFormData =
```
lucca65, line 115: Much better naming!

src/elm/Auth.elm
```diff
@@ -189,81 +274,182 @@ view isModal shared model =
...
+        ]
+        ([ viewAuthError shared model.loginError ]
+            ++ (case loginStep of
+                    LoginStepPassphrase ->
+                        viewPassphrase
+
+                    LoginStepPIN ->
+                        viewCreatePin
+               )
+        )
```
lucca65, line 448: why not this:

```elm
[ div [ . . . ]
        [ viewAuthError shared model.loginError
        , case loginSetup of
             . . .
]

File: src/elm/Auth.elm [hunk 22]
```diff
@@ -349,102 +519,39 @@ viewAuthError shared maybeLoginError =
...
-    in
-    div [ class "card__auth__pin__section" ]
-        [ viewFieldLabel shared "auth.pin" "pin_input_0" (Just (toggleViewPin model))
-        , div [] inputs
+toggleViewPin : Bool -> String -> String -> Msg -> Html Msg
+toggleViewPin isVisible showLabel hideLabel msg =
+    button
+        [ class "absolute mt-3 uppercase text-xs right-0 mr-3 text-orange-300"
+        , onClick msg
+        , attribute "tabindex" "-1"
```
lucca65, line 532: nice!! we should have added this long ago!

src/elm/Auth.elm
```diff
@@ -584,27 +771,26 @@ update msg shared model showAuthModal =
                 }
 
         SubmittedLoginPIN ->
-            if List.any (\a -> a == Nothing) model.form.enteredPin then
-                { model | loginError = Just "Please fill in all the PIN digits" }
-                    |> UR.init
+            case model.form.enteredPin of
```
lucca65, line 774: maybe we should use use `if String.isEmpty model.form.enteredPin then`

src/elm/Auth.elm
```diff
@@ -804,11 +962,145 @@ msgToString msg =
...
+        KeyPressed _ ->
+            [ "KeyPressed" ]
+
+
+viewPin : Model -> Shared -> Html Msg
+viewPin ({ form, problems } as model) shared =
+    let
+        pinLabel =
+            case shared.maybeAccount of
+                Nothing ->
```
lucca65, line 986: I would invert this, looks weird to look for `Nothing` instead of `Just`

src/elm/Auth.elm
```diff
@@ -804,11 +962,145 @@ msgToString msg =
...
+            I18Next.t shared.translations
+
+        tr : String -> I18Next.Replacements -> String
+        tr =
+            I18Next.tr shared.translations I18Next.Curly
+
+        inputType =
+            if isVisible then
+                -- `"text"` is used here because with `"number"` field restrictions for the PIN
+                -- don't apply after toggling visibility (see `trimPinNumber` function).
```
lucca65, line 1066: +1

src/elm/Page/Register.elm
amiskov: I think it's fine here.

The thing is that we should have the name in bold:
image

So we have to separate `Hey`, `name`, and other text. Using the space at the end (`"Hey "`) is error-prone. The comma shouldn't be bold, but using it as the first symbol (`", it is the last..."`) looks odd in translations. So I decided to make the separation very obvious. It may look strange in Elm, but it's more convenient for translations.

src/elm/Page/Register.elm
amiskov: Found out that we already have `left-1` class :)

src/elm/Page/Register.elm
amiskov: Actually `"max-width" "10rem"` is redundant here. Removed it.

src/elm/Auth.elm
amiskov: In this case, we would have different types in a list:

- `viewAuthError ...` is `Html Msg`,
- but `case loginSetup of...` is `List (Html Msg)`.

Anyway, this code looks complicated, I made it more readable.

src/styles/main.css
amiskov: I haven't found anything similar to [this behavior](https://codepen.io/chriscoyier/pen/RRbKrL) in existing styles. Maybe I missed something?

These styles are just a little abstraction above several flex-related styles. I decided to create them because it was more convenient to use meaningful class names in a mess of HTML elements among all other flex-related classes.

amiskov:
@lucca65 Thank you for this thoughtful review! There are only a couple of your comments that I didn't resolve, everything else I've fixed. Please, have a look!

src/elm/Page/Register.elm
amiskov: We could've used interpolation like `Hey <b>{{username}}</b>, ...` in the translation template, but we can't use HTML tags in there...

src/styles/main.css
lucca65: We do, our footer uses `flex-grow` to achieve this: https://github.com/cambiatus/frontend/blob/b62fdedc696f8bc597c453296ca1c98c1bf872c6/src/elm/Session/LoggedIn.elm#L288

PR #256, opened 2020-05-27
Only accept an odd number of validators 

### LINKED ISSUE #243: Make the validator # to be at minimum 3 on Action Editor
**Describe the bug**
Right now we require a minimum of 2 votes to create a new action (on the `ActionEditor.elm` file). The problem is that this causes a scenario where claims that require two votes (actually all scenario with an even number of required votes) we may reach a scenario of having half of the required votes as "approved" and "disapproved". This will create limbo scenario of ever pending claims. 

**To Reproduce**
1. Create action with a number of two verifications (may have several more validators on the list)
2. Open a new claim
3. As the validators, vote positive as the first validator and negative as the second validator
4. See the zombie claim (never approved or disapproved)

**Expected behavior**
We should be able to make sure the claim reaches the end of its cycle. To make sure it will happen we should make sure we only accept odd numbers (but bigger than 1)

RodrigoLeiteF:
Closes #243 

Prevents an even number of validators when creating a new action

### DIFF
--- a/public/translations/pt-BR.json
+++ b/public/translations/pt-BR.json
@@ -56,7 +56,7 @@
         "lower_than_or_equal": "O número deve ser menor ou igual a {{base}}",
         "is_odd": "O número deve ser ímpar"
       },
-      "DATE": {
+      "date": {
         "invalid": "A data está inválida. Use 'mm/dd/aaaa'"
       },
       "text": {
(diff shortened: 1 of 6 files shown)


public/translations/pt-BR.json
```diff
@@ -53,9 +53,10 @@
         "greater_than_or_equal": "O número deve ser maior ou igual a {{base}}",
         "length_greater_than_or_equal": "Escolha ao menos {{count}} itens",
         "lower_than": "O número deve ser menor do que {{base}}",
-        "lower_than_or_equal": "O número deve ser menor ou igual a {{base}}"
+        "lower_than_or_equal": "O número deve ser menor ou igual a {{base}}",
+        "is_odd": "O número deve ser ímpar"
       },
-      "date": {
+      "DATE": {
```
lucca65, line 59: why the uppercase?

RodrigoLeiteF: Oops

PR #262, opened 2020-06-02
Do project wide warning cleanup

lucca65:
No issue is related to this PR

Do a cleanup to the project warning, since we moved it from the private repository we had lots of old warnings, refactors that still used some of the old pages structure, etc

This intend do to a clean up of several warnings, but not all, some of them are still there:

- A few from our auto generated Graphql types, those will be automatically generated/updated as we add stuff to the backend api, so no point in fixing them
- A query from transfer, that may be more useful to remove after we added the Timeline feature
- a few TODOs that are still relevant somehow

The test is just to make sure the project compiles properly, no behaviours were changed

### DIFF
--- a/src/elm/Session/LoggedIn.elm
+++ b/src/elm/Session/LoggedIn.elm
@@ -40,7 +40,7 @@ import Flags exposing (Flags)
 import Graphql.Document
 import Graphql.Http
 import Graphql.Operation exposing (RootSubscription)
-import Graphql.SelectionSet as SelectionSet exposing (SelectionSet, with)
+import Graphql.SelectionSet exposing (SelectionSet)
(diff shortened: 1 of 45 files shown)


src/elm/Session/LoggedIn.elm
```diff
@@ -967,13 +967,16 @@ isAccount accountName model =
 
 
 type alias UnreadMeta =
-    { unreads : Int }
+    Int
+
+
+
+-- { unreads : Int }
```
amiskov, line 974: Does this comment mean something?

lucca65: nice catch! fixed!

PR #269, opened 2020-06-12
Add page titles

### LINKED ISSUE #254: Add page titles for better navigation
All pages have the same title:
cambiatus_page_titles

It's better to have a meaningful title for each page to be able to navigate via browser history.

Change page titles along with URI during the routing.

**Additional context**

amiskov:
Closes #254.

- [x] Make sure that all `<SomePage>.view` functions return the value of type ~~[`Document Msg`](https://package.elm-lang.org/packages/elm/browser/latest/Browser#Document)~~ `{ title : String, content : Html Msg }` instead of `Html Msg`.
- [x] Add translations for all the titles.

- Every page should have its own title in the browser tab.
- The format should be like this: `<Some Page Title> | Cambiatus`.


src/elm/Page/Community.elm
```diff
@@ -112,7 +112,7 @@ type alias Member =
 -- VIEW
 
 
-view : LoggedIn.Model -> Model -> Html Msg
+view : LoggedIn.Model -> Model -> { title : String, content : Html Msg }
```
lucca65, line 115: maybe we should use a type alias for this, and the best place to add it is “Page.elm” what do you think @amiskov?

amiskov: I knew that you will ask about this :) I'll try to explain why I decided not to create a dedicated type alias.

In Elm, we already have very similar [`Browser.Document`](https://package.elm-lang.org/packages/elm/browser/latest/Browser#Document) type for this but with `List (Html msg)` for HTML output:

```elm
type alias Document msg =
    { title : String
    , body: List (Html msg)
    }
```

I actually [tried](https://github.com/cambiatus/frontend/pull/269/commits/26359e2fc82434ae2f8367624f80cf3c9bcf7824) to use `Document` from all `view` functions, but I didn't like [the way it worked](https://github.com/cambiatus/frontend/pull/269/commits/26359e2fc82434ae2f8367624f80cf3c9bcf7824#diff-6e637b873a9a45f19ff29facd323d55dR960) (line 960) with existing code in `Main.elm`, where `Html msg` was used instead of `List (Html msg)`. I also tried to change everything to work with `List (Html msg)` but it was just too much to refactor.

I decided not to create a dedicated type because it would be very similar to `Browser.Document`. I used just a record to indicate that this is intentionally Ad Hoc solution, which will become `Document` later, in `Main.view`. I found that this kind of implementation is used in [`elm-spa-example`](https://github.com/rtfeldman/elm-spa-example/blob/master/src/Page/Profile.elm#L145) and it looked pretty good for me.

lucca65: I see @amiskov, thanks a lot for the explanation!

Issue #266, opened 2020-06-09
Create a generic PDF generator app


lucca65:
**Is your feature request related to a problem? Please describe.**
As we migrated to using only 12 words as login options, some old users that used the private key to login are having issues to use the app. 

Right now I'm manually generating new PDFs for those users and changing their keys. Since this is happening a lot lately, I think it would be good to have a tool to automate that process

**Describe the solution you'd like**
Maybe a simple, GitHub hosted static page that simply generate a new random PDF with the public key, this way we can just setup the new key more easily

**Describe alternatives you've considered**
I think this is the best option other than just keep doing it manually

**Additional context**

lucca65:
after a few months with this issue opened I no longer think this is necessary. it would help during the migration for users that had the old PDF but it's a very limited number of users

PR #270, opened 2020-06-18
New profile page

### LINKED ISSUE #258: My Profile Refactor

RodrigoLeiteF:
Closes #258 

Refactors profile page to be aligned with the new design

Visit your profile by hovering your profile picture and clicking Profile

The new design shares a lot of it's design with the public profile page. We should probably look into merging them on a single page, as to reduce repetitiveness.


amiskov:
A few words about the generation of the PDF file with the passphrase (12 words) introduced by 49860f0.

There's one general function for generating PDF, but there are 2 different ports to run it:

- `downloadAuthPdfFromRegistration`;
- and `downloadAuthPdfFromProfile`.

I introduced 2 ports because we have to gather needed data (12 words and the account name) from different places:

- for the Registration page - from the form filled by the user;
- for the Profile page - from the `localStorage`, where encrypted data is stored.

I mentioned the account name because I changed the filename of the PDF from `Cambiatus.pdf` to `<accountName>_cambiatus.pdf`. I think it will help users keep things more organized (e.g. differ PDFs for the different accounts or just have more info about what's in PDF without opening it).

lucca65:
nice one on the PDF name

amiskov:
I found out that we need to update the Login with PIN modal dialogue to be arranged with the current design. Also, the text may be improved as well.

This dialogue is shown only for already signed-in users when we want the PIN from them to confirm some operation: transfer, changing profile, etc. So the “Continue” will work better than “Sign-in” for the button text. We also don’t need to welcome the user since he/she is already signed-in.
new-login-with-pin-modal
I already made a commit so you can check the code and leave your notes in there: https://github.com/cambiatus/frontend/pull/270/commits/eb94e072bb65c21c47cefba2a0034d5d0b8c2a73

@heltonlr @lucca65 are you ok with that changes?

heltonlr:
Amazing @amiskov. Only check the weight of the button text, we must use bold.

amiskov:
That's a great catch, @lucca65! Thanks, I really didn't notice these problems...

> it doesn't work on smaller screens (320px), it gets cropped.

My bad. Didn't check that size.

> My recommendation is quit using grid to keep the whole project consistent.

I agree. Grids are nice to use in overall layout framing but in smaller details like the forms, they lack flexibility and became too complex. It's hard to maintain and fix design issues in the grid-based components because the elements are coupled with each other. For example, in this grid every row becomes locked with the 3-cols layout, which is relevant only for the 1st row with the userpic and the name:
image

@RodrigoLeiteF hope you will agree with that!

> I'm also getting an error when trying to download the PDF

I found the cause of this problem. In this branch, I introduced the `.encryptedPassphrase` field in the user data stored in the `localStorage` and this field is available only if you are logged in using the code from this branch. When you were logged in using the `master` branch and then switch to this PR's branch then this field will not be available. Actually, any user could have this error. I'm on my way to figure out the solution...

amiskov:
### Download PDF, the things to note

**Already logged in users will be able to download PDF with the passphrase only after re-login**.

That's because in the current setup we don't store the passphrase (12 words) in the browser, so we can't retrieve it to show in PDF <sup>*</sup>. We only store the encrypted private key. The #266 issue is related to this case.

This PR introduces the ability to store the encrypted passphrase in the user's browser during the login.

My only suggestion, for now, is to show the message for the user saying that we don't know 12 words but we can help with getting the new ones.

How to reproduce the case when there's no passphrase:

- in your current branch (or in `master`) run `yarn start` and log in;
- in another terminal switch to `feature/new-profile-page`, make sure the build is recompiled;
- go to the Profile and try to download PDF. This message should appear:
image

How to reproduce the OK behavior:

- checkout the `feature/new-profile-page` branch, log out and then log in;
- go to the Profile page and try to download the PDF; it should be downloaded properly with the passphrase.

---

<sup>*</sup> Well, we actually can in the very limited case when the user didn't reload the page after logging in. It works because the passphrase is stored in the Elm model until the page is reloaded. It's more like an unintentional feature and we can't rely on it (the [code](https://github.com/cambiatus/frontend/blob/master/src/index.js#L388)).

RodrigoLeiteF:
I can't review my own pull request and there's no way to transfer it over to you, but everything looks great from what I can see. Good job :)

src/elm/Page/Profile.elm
```diff
@@ -121,452 +105,342 @@ view loggedIn model =
...
+
+                Just pk ->
+                    pk
+
+        content =
+            div
+                []
+                [ p
+                    [ class "w-full font-medium text-heading text-2xl mb-2" ]
+                    -- I'll add the translations after we agreed about this case
```
lucca65, line 321: I'm 100% on this. This corner case happened a few times on MUDA and this will be good solution since a few users simply stopped using after facing this

src/elm/Page/Profile.elm
```diff
@@ -121,452 +105,342 @@ view loggedIn model =
...
+                , p [ class "font-bold my-3 text-lg text-center border p-4 rounded-sm bg-gray-100" ]
+                    [ text privateKey
+                    ]
                 ]
-            , div [ class "profile-badges-item" ]
-                [ img [ src "images/badges-mentor.svg" ]
-                    []
-                , span [] [ text_ "profile.badge.mentor" ]
+    in
+    Modal.view
```
lucca65, line 335: NICE!

src/elm/View/Modal.elm
```diff
@@ -0,0 +1,59 @@
...
+                [ class "modal container fade-in"
+                , stopPropagationOn "click" (Decode.succeed ( cfg.ignoreMsg, True ))
+                ]
+                [ div
+                    [ class "modal-bg"
+                    , onClick cfg.closeMsg
+                    ]
+                    []
+                , div
+                    [ class "modal-content overflow-auto" ]
```
lucca65, line 46: maybe this `overflow-auto` is messing up the content when the error message appears, check it out:

Screen Shot 2020-07-15 at 14 23 05

src/elm/View/Pin.elm
```diff
@@ -0,0 +1,109 @@
...
+
+
+isValid : Pin -> Bool
+isValid pin =
+    let
+        hasCorrectLength p =
+            String.length p == 6
+
+        hasOnlyDigits p =
+            String.all Char.isDigit p
```
lucca65, line 32: no need to add p here, it can be omitted, currying will make it work!

src/elm/View/Pin.elm
```diff
@@ -0,0 +1,109 @@
+module View.Pin exposing (..)
+
+import Html exposing (..)
+import Html.Attributes exposing (..)
```
lucca65, line 4: elm-review will probably complain about this `exposing (..)`

src/elm/View/Modal.elm
```diff
@@ -0,0 +1,59 @@
+module View.Modal exposing (..)
```
lucca65, line 1: elm-review will probably complain about this

src/elm/View/Modal.elm
```diff
@@ -0,0 +1,59 @@
...
+    Shown
+
+
+view : Visibility -> Config msg -> Html msg
+view status cfg =
+    case status of
+        Shown ->
+            div
+                [ class "modal container fade-in"
+                , stopPropagationOn "click" (Decode.succeed ( cfg.ignoreMsg, True ))
```
lucca65, line 38: nice!

src/index.js
```diff
@@ -192,6 +197,21 @@ function getSelectedCommunity () {
   return window.localStorage.getItem(SELECTED_COMMUNITY_KEY)
 }
 
+function downloadPdf (accountName, passphrase, responseAddress, responseData) {
+  const definition = pdfDefinition(passphrase)
+  const pdf = pdfMake.createPdf(definition)
+
+  pdf.download(accountName + '_cambiatus.pdf')
```
lucca65, line 204: +1

src/elm/Page/Profile/Editor.elm
```diff
@@ -0,0 +1,476 @@
...
+                Loaded profile ->
+                    view_ loggedIn model profile
+    in
+    { title = title
+    , content = content
+    }
+
+
+view_ : Session.LoggedIn.Model -> Model -> Profile -> Html Msg
+view_ loggedIn model profile =
```
lucca65, line 104: something is off here. checkout how the font size looks bigger:

Screen Shot 2020-07-15 at 14 33 07
Screen Shot 2020-07-15 at 14 33 02

src/elm/Icons.elm
```diff
@@ -132,6 +177,18 @@ trash class_ =
     svg [ width "24", height "24", viewBox "0 0 24 24", class class_ ] [ Svg.path [ fillRule "evenodd", clipRule "evenOdd", d "M20.3077 5.16788H16.3477C15.6644 3.41192 13.9331 2.25 12 2.25C10.0669 2.25 8.33558 3.41192 7.65231 5.16788H3.69231C3.30996 5.16788 3 5.46885 3 5.84013C3 6.2114 3.30996 6.51237 3.69231 6.51237H4.15385V20.4055C4.15385 21.148 4.77376 21.75 5.53846 21.75H18C18.7647 21.75 19.3846 21.148 19.3846 20.4055V6.51237H20.3077C20.69 6.51237 21 6.2114 21 5.84013C21 5.46885 20.69 5.16788 20.3077 5.16788ZM12 3.5993C13.26 3.5993 14.3538 4.04746 14.8938 5.16788H9.10615C9.64615 4.04298 10.74 3.5993 12 3.5993ZM18 20.4055H16.1538V12.7867C16.1538 12.4154 15.8439 12.1144 15.4615 12.1144C15.0792 12.1144 14.7692 12.4154 14.7692 12.7867V20.4055H12.4615V10.0977C12.4615 9.72643 12.1516 9.42545 11.7692 9.42545C11.3869 9.42545 11.0769 9.72643 11.0769 10.0977V20.4055H8.76923V12.7867C8.76923 12.4154 8.45927 12.1144 8.07692 12.1144C7.69457 12.1144 7.38462 12.4154 7.38462 12.7867V20.4055H5.53846V6.51237H18V20.4055Z", fill "#DB1B1B" ] [] ]
 
 
+edit : String -> Svg msg
+edit class_ =
```
lucca65, line 181: here `class_` is not used. 

Just to give some context, we use this param to allow us to add color to icons.

src/elm/Page/Profile.elm
```diff
@@ -971,23 +707,22 @@ jsAddressToMsg addr val =
                     -- TODO: Handle PushSubscription Decode error
```
lucca65, line 707: I know it was already here, but can you remove this TODO @amiskov?

This is related to notifications but our context changed a lot since we added this, moving forward we should add a more intelligent notification system on the backend, allowing for emails and push.

For now we can just remove it

src/elm/Page/Profile.elm
```diff
@@ -997,68 +732,47 @@ msgToString msg =
...
-
-        EnteredEmail _ ->
-            [ "EnteredEmail" ]
-
-        EnteredBio _ ->
-            [ "EnteredBio" ]
-
-        EnteredLocalization _ ->
-            [ "EnteredLocalization" ]
+        DownloadPdf r ->
```
lucca65, line 735: r is unused here

src/elm/Page/Profile.elm
```diff
@@ -997,68 +732,47 @@ msgToString msg =
...
-            [ "ClickedViewPrivateKey" ]
+        TogglePinReadability ->
+            [ "TogglePinReadability" ]
+
+        ChangePinSubmitted ->
+            [ "ChangePinSubmitted" ]
 
-        ClickedClosePrivateKey ->
-            [ "ClickedClosePrivateKey" ]
+        EnteredPin r ->
```
lucca65, line 762: `r` is unused here

src/elm/Page/Profile.elm
```diff
@@ -997,68 +732,47 @@ msgToString msg =
...
+            [ "EnteredPin" ]
+
+        GotPushPreference r ->
```
lucca65, line 765: `r` is unused

src/elm/Page/Profile/Editor.elm
```diff
@@ -0,0 +1,476 @@
...
+            ]
+        ]
+
+
+
+-- UPDATE
+
+
+type Msg
+    = Ignored
```
lucca65, line 276: not used

src/elm/View/Modal.elm
lucca65: oh I forgot you created an issue for this. #292 should address this!

src/elm/View/Pin.elm
amiskov: Good point! As for the point-free :)

src/elm/Page/Profile.elm
amiskov: Done.

src/elm/View/Pin.elm
amiskov: Now we haven't this rule enabled. Just added it to use in the future. I tried it on the whole project and it worked well since we don't expose everything anywhere else.

src/elm/Page/Profile/Editor.elm
amiskov: Good catch! I rechecked the [design in Figma](https://www.figma.com/file/oZJUIyHRmLyLIKDxYebvLg/Form-elements?node-id=0%3A1) for the form elements and improved CSS rules a bit.

amiskov:
> What do you think about working on this PR @amiskov? It turned out to be a pretty big change

@lucca65, Since @RodrigoLeiteF did a pretty big part of the job on this PR the changes were not very big for me :) I mainly worked on the PIN updating, downloading the PDF, and polishing the design.

src/elm/Page/Profile.elm
amiskov: The text in the popup says that the users could contact us. How they can do that? Do we have any email which could be used for the user's support?

lucca65: lets discuss this over on #300

lucca65: for this PR alone I think just the text is enough. Thoughts @heltonlr

src/elm/View/Pin.elm
lucca65: oh I see! good one @amiskov

PR #308, opened 2020-07-26
Put back the objectives that were removed

lucca65:
N/A

Put back the objectives section that was removed by accident. It caused a lot of trouble to MUDA, where no user was able to open new claims. We should be more careful on next reviews and don't remove nothing that we are not sure that needs to be removed or not

Open a community that has objectives enabled and filled in, and you should see something like this:
Screen Shot 2020-07-25 at 22 39 33

### DIFF
--- a/src/elm/Page/Community.elm
+++ b/src/elm/Page/Community.elm
@@ -157,7 +157,7 @@ view loggedIn model =
                                 ]
                             , p [ class "text-grey-200 text-sm" ] [ text community.description ]
                             ]
-                        , if canEdit && community.hasObjectives then
+                        , if community.hasObjectives then
                             div [ class "container mx-auto px-4" ]
                                 [ div [ class "bg-white py-6 sm:py-8 px-3 sm:px-6 rounded-lg mt-4" ]
                                     (Page.viewTitle (t "community.objectives.title_plural")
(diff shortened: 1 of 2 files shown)


src/elm/Page/Community.elm
```diff
@@ -173,7 +173,17 @@ view loggedIn model =
...
+                                 (Page.viewTitle (t "community.objectives.title_plural")
+                                     :: List.indexedMap (viewObjective loggedIn model community)
+                                         community.objectives
+                                     ++ [ if canEdit then
+                                             viewObjectiveNew loggedIn editStatus community.symbol
+
+                                           else
+                                             text ""
+                                        ]
+                                 )
```
amiskov, line 186: The `else` branch is the same as `then`. And the `div [ class "container mx-auto px-4" ]`, like in the `then` branch, is also important here. It makes the block to be rounded and restricts the `width` of the block.

It looks like the cause of the problem is the `canEdit` predicate. I guess it's better to just remove it and don't duplicate the code:

```elm
                        , if community.hasObjectives then
                            div [ class "container mx-auto px-4" ]
                                [ div [ class "bg-white py-6 sm:py-8 px-3 sm:px-6 rounded-lg mt-4" ]
                                    (Page.viewTitle (t "community.objectives.title_plural")
                                        :: List.indexedMap (viewObjective loggedIn model community)
                                            community.objectives
                                        ++ [ if canEdit then
                                                viewObjectiveNew loggedIn editStatus community.symbol

                                             else
                                                text ""
                                           ]
                                    )
                                ]

                          else
                            text ""
                        ]
```

@RodrigoLeiteF Please, have a look!

lucca65: You are right @amiskov, I missed the point here

PR #325, opened 2020-08-03
Remove all references to IPFS

### LINKED ISSUE #282: Move away from IPFS
After done in the backend, we need to update the URL on the config so the images keep working

RodrigoLeiteF:
Closes #282 

Removes references to IPFS and adapts the platform to use the new upload system

- [ ] All image references use the value they receive from the API directly, without appending the IPFS base URL
- [ ] Uploading an avatar uses the endpoint `/upload` instead of `/ipfs`


src/elm/Community.elm
```diff
@@ -232,29 +232,24 @@ settingsQuery symbol =
...
+logoBackground : Maybe String -> Html.Attribute msg
+logoBackground maybeHash =
     Html.Attributes.style "background-image"
-        ("url(" ++ logoUrl ipfsUrl maybeHash ++ ")")
-
-
-logoPlaceholder : String -> String
-logoPlaceholder ipfsUrl =
-    ipfsUrl ++ "/QmXuf6y8TMGRN96HZEy86c8N9aDseaeyuCQ5qVLqPyd8Ld"
+        ("url(" ++ logoUrl maybeHash ++ ")")
```
RodrigoLeiteF, line 252: Any ideas on how we should handle this placeholder? Maybe we could have the API return the information, and we'd fetch it once and cache. If we're going to hardcode a URL it might be better to just serve it as a static asset in the frontend itself (maybe as an SVG?)

lucca65: +1 to adding it to the project image folder

src/elm/Page/Community/Editor.elm
```diff
@@ -429,7 +426,7 @@ viewFieldLogo shared isDisabled selected logos errors =
                     Uploaded hash ->
```
lucca65, line 426: good opportunity to change `hash` to `url` or something similar

src/elm/Page/Shop/Editor.elm
```diff
@@ -250,7 +247,7 @@ viewForm shared balances imageStatus isEdit isDisabled deleteModal form =
         imageStyle =
             case getInput form.image of
                 Just hash ->
```
lucca65, line 249: good place to change the term `hash`

src/elm/Community.elm
RodrigoLeiteF: Done. I can't find a place to test it, but I followed the way other icons are shown in the application so it should work just fine :)

public/icons/shop-placeholder.svg
```diff
@@ -0,0 +1,19 @@
...
+<path d="M204.818 97.2027L194.902 112.929C193.979 114.381 193.691 116.075 193.979 117.768C194.325 119.462 195.248 120.914 196.631 121.881C198.015 122.849 199.629 123.152 201.244 122.849C202.858 122.486 204.242 121.519 205.164 120.067L215.08 104.34C216.925 101.376 216.176 97.3237 213.351 95.3881C212.313 94.6622 211.16 94.2993 209.949 94.2993C209.546 94.2993 209.142 94.3598 208.681 94.4203C207.067 94.7832 205.683 95.751 204.818 97.2027Z" fill="white"/>
+<path d="M204.876 104.34L214.792 120.067C215.715 121.519 217.098 122.486 218.713 122.849C220.327 123.212 221.999 122.849 223.325 121.882C226.15 119.885 226.957 115.893 225.054 112.929L215.138 97.2027C214.216 95.751 212.832 94.7832 211.218 94.4203C210.814 94.3598 210.411 94.2993 209.949 94.2993C208.739 94.2993 207.585 94.6622 206.548 95.3881C203.78 97.3842 203.031 101.376 204.876 104.34Z" fill="white"/>
```
lucca65, line 18: looks like the migration worked but there is only one placeholder image, maybe we can use all they provided! should be quick to choose a random image from all the placeholders!

src/elm/Page/Shop.elm
```diff
@@ -254,25 +255,22 @@ viewShopFilter loggedIn filter =
 viewGrid : LoggedIn.Model -> List Card -> Model -> Html Msg
 viewGrid loggedIn cards model =
     let
-        v_ viewFn card =
-            viewFn model loggedIn card
+        v_ viewFn index card =
+            viewFn model loggedIn card index
```
amiskov, line 259: This `v_` was quite confusing for me... Maybe we could make it simpler?

If we switch `Int` and `Card` arguments in `viewCard` function then currying will do the work:

```elm
viewGrid : LoggedIn.Model -> List Card -> Model -> Html Msg
viewGrid loggedIn cards model =
    div [ class "flex flex-wrap -mx-2" ]
        (List.indexedMap
            (viewCard model loggedIn)
            cards
        )


viewCard : Model -> LoggedIn.Model -> Int -> Card -> Html Msg
viewCard model ({ shared } as loggedIn) index card =
    -- ...
```

src/elm/Page/Shop.elm
```diff
@@ -303,7 +301,7 @@ viewCard model ({ shared } as loggedIn) card =
         ]
         [ div [ class "md:hidden rounded-lg bg-white h-32 flex" ]
             [ div [ class "w-1/4" ]
-                [ img [ class "rounded-l-lg object-cover h-32 w-full", src imageUrl ] []
+                [ img [ class "rounded-l-lg object-cover h-32 w-full", src image, on "error" (Json.Decode.succeed (OnImageError index)) ] []
```
amiskov, line 304: Didn't know that we can handle image source loading errors this way 👍

src/elm/Page/Shop.elm
```diff
@@ -510,6 +509,36 @@ update msg model loggedIn =
...
+                                newCard =
+                                    { card | sale = newSale }
+
+                                newList =
+                                    Array.set index newCard cardArray
+                            in
+                            { model | cards = Loaded (Array.toList newList) } |> UR.init
+
+                        Nothing ->
+                            UR.logImpossible msg [ "cardOutOfIndex" ] (UR.init model)
```
amiskov, line 537: One more thing I learned from this PR is logging errors in production :) 👍

src/elm/Page/Shop.elm
RodrigoLeiteF: Good catch, I didn't know we could do this :)

src/elm/Page/Shop.elm
RodrigoLeiteF: This was news for me as well. I was hoping there would be an HTML-only way, but apparently the only one that exists involves a bit of hacking with the `object` tag. This is the next best thing :p

lucca65: wow this is actually pretty neat

public/icons/shop-placeholder.svg
RodrigoLeiteF: Instead of using a random seed, I went with a simpler deciding method that simply uses the index of the card. Randomness in Elm is a bit annoying, and this has a similar effect while being simpler. Hope it's ok :)

lucca65: I think it is! very elegant solution

PR #331, opened 2020-08-06
Fix/claim modal

lucca65:
None, but we had reports from users not being able to claim nothing.

Add back the claim modal into the community page. I've also started using the new modal pattern, much better!

Try to claim an action on the community page!

### DIFF
--- a/src/elm/Page/Community.elm
+++ b/src/elm/Page/Community.elm
@@ -33,6 +33,7 @@ import Task
 import Time exposing (Posix, posixToMillis)
 import UpdateResult as UR
 import Utils
+import View.Modal as Modal
 
 
 
@@ -159,7 +160,8 @@ view loggedIn model =
                             ]
                         , if community.hasObjectives then
                             div [ class "container mx-auto px-4" ]
-                                [ div [ class "bg-white py-6 sm:py-8 px-3 sm:px-6 rounded-lg mt-4" ]
+                                [ viewClaimModal loggedIn model
+                                , div [ class "bg-white py-6 sm:py-8 px-3 sm:px-6 rounded-lg mt-4" ]
                                     (Page.viewTitle (t "community.objectives.title_plural")
                                         :: List.indexedMap (viewObjective loggedIn model community)
                                             community.objectives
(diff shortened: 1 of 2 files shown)


src/elm/Session/LoggedIn.elm
```diff
@@ -160,7 +160,7 @@ initModel shared authModel accountName selectedCommunity =
     , auth = authModel
     , feedback = Hidden
     , showCommunitySelector = False
-    , hasShop = False
```
lucca65, line 163: I've added this from the other PR #330 so I could do a deploy for MUDA (no one is being able to claim actions!!)

src/elm/Page/Community.elm
```diff
@@ -482,6 +484,51 @@ viewAction loggedIn metadata maybeDate action =
...
+                            , if not isLoading then
+                                onClick CloseClaimConfirmation
+
+                              else
+                                onClick NoOp
+                            , disabled isLoading
+                            ]
+                            [ text_ "dashboard.check_claim.no" ]
+                        , button
+                            [ class "flex-1 block button button-primary button-lg w-full md:w-40"
```
amiskov, line 515: There are predefined classes for this kind of buttons from the modal's footer: `modal-cancel` and `modal-accept`. Please, see [`src/elm/Page/Dashboard.elm:225`](https://github.com/cambiatus/frontend/blob/master/src/elm/Page/Dashboard.elm#L225) for the reference (the claim popup on the Dashboard).

src/elm/Page/Community.elm
```diff
@@ -482,6 +484,51 @@ viewAction loggedIn metadata maybeDate action =
             ]
 
 
+viewClaimModal : LoggedIn.Model -> Model -> Html Msg
+viewClaimModal loggedIn model =
+    case model.modalStatus of
+        Opened isLoading actionId ->
```
RodrigoLeiteF, line 490: Just a small thing. This should be `open` instead. `Opened` is used as the past tense of `open` (he opened the door). We use `open` when we want to denote `not closed`.

src/elm/Page/Community.elm
lucca65: Nice! thanks for the review and pointing out an example! It helped a lot 💪

src/elm/Page/Community.elm
lucca65: nice! I'll fix it! small change but important

PR #337, opened 2020-08-13
Use S3 URLs for default community icons

RodrigoLeiteF:
----

Closes https://cambiatus.slack.com/archives/CA83HJAAD/p1597298920267800

Default community icons now use S3 URLs

Access http://localhost:4000/community/new


RodrigoLeiteF:
> I think I'm ok with that for now!
> The only caveat I can imagine is that we may forget to change this URL if we decide to change the host for images once again :)

Agreed. We should probably add a BASE_URL setting in the future like we did for IPFS in the future.

Issue #312, opened 2020-07-27
Change invitation flow for existing users 15


lucca65:
Users that are invited to a KYC enabled community must fill in more data when they are invited. The KYC is related to the user but we need to make sure we have everything during the community invitation process.

Related [Figma designs](https://www.figma.com/file/47BEsGEMSWfhTm3zC0ugyA/Main-flow?node-id=3%3A2206) with all the fields the user should have filled in to join the community with KYC enabled.

amiskov:
Here's how I see the flow:
![image](https://user-images.githubusercontent.com/140053/90560888-7e6c3f00-e1a8-11ea-9f68-eb542943e44b.png)

@lucca65 do I understand everything right?

How do you think, should we allow the existing users to accept invitations and become a company (show 2 additional screens with toggle, like during the registration)? It doesn't seem reasonable (existing users are not companies), but maybe I don't know all the details...

lucca65:
>  It doesn't seem reasonable 

+1. The way I imagined, we would add the migration for the user type, defaulting every existing user to the `natural`  type. Then the invitation for existing users would just ask for the kyc information if needed, if not, it just adds to the community like normal

amiskov:
Here's the invitation process flow for the existing user as I think of it:

![image](https://user-images.githubusercontent.com/140053/90630063-0560fc00-e229-11ea-83a2-f6119034355b.png)

amiskov:
It looks like we need the backend part and the mutations for saving, updating, and deleting the KYC user data.

At first glance, it could be like this:

- create `saveKycData` mutation for saving the KYC data:
    + use it during the registration (#311 issue);
    + use it during the invitation of the existing user to the KYC-enabled community (this issue) for the user who hasn't KYC fields field;
- `updateKycData(account_id)` for updating user's KYC data from the profile page;
- `deleteKycData(account_id)` to delete the user's KYC info if he/she wants to do so.

Here are my concerns about that:

- we already have `updateProfile` mutation and it looks reasonable to use it also for updating the KYC fields from the Profile page. I see it like this: send all the user profile fields (including KYC) with `updateProfile` mutation and handle the data update on the backend (update KYC parts in the `kyc_table`, save user data to the `users` table).
- we already have `signUp` mutation and it is also reasonable to send all the data from Register/Invite page to it and handle saving to different tables on the backend.

So, here's my proposal:

[...]

Issue #311, opened 2020-07-27
Update Sign Up process to use the new GraphQL mutation and new forms. 40


lucca65:
It will only be available through invitations and it need to know if the community has enabled KYC in order to render the full form

RodrigoLeiteF:
Fixed by #341

PR #339, opened 2020-08-19
Improve invitation process for inviter and invitee

amiskov:
Closes N/A. Contains handy changes for the invitation process; related to #312 and #311.

- Show notice if the invitee is already a member of the inviter's community;
- Fix the invite link that it may contain a port number like `:3000` (we are doing a lot of tests on localhost so it will help to save time);
- Improve code readability.

image

- Go to Dashboard and press "Invite your friends" on your localhost. There should be a link with a port `:3000`.
- Pretend you are an invitee and open the link. There should be a notice that you are already a member of this community.
- Open the link logged-out and see the usual accept the invitation page.
- Open the link from the user account which is not a member of this community, see the confirmation page (same as for logged-out user).

### DIFF
--- a/src/elm/Page/Community/Invite.elm
+++ b/src/elm/Page/Community/Invite.elm
@@ -16,42 +16,24 @@ import Html exposing (Html, button, div, img, span, text)
 import Html.Attributes exposing (class, src)
 import Html.Events exposing (onClick)
 import Http
-import I18Next exposing (t)
 import Page exposing (Session(..), toShared)
 import Profile exposing (Profile)
 import Route
 import Session.LoggedIn as LoggedIn exposing (External)
-import Session.Shared exposing (Shared)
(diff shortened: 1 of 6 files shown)


src/elm/Page/Community/Invite.elm
```diff
@@ -64,54 +47,112 @@ type ModalStatus
...
-                                ]
-
-                        Error e ->
-                            Page.fullPageError (t shared.translations "") e
-                    ]
-                , viewFooter session
-                ]
+            case model.pageStatus of
+                Loading ->
+                    div [] []
```
lucca65, line 116: since we are modifying this, maybe we can just use 

```
text ""
```

amiskov: Done!

PR #340, opened 2020-08-20
Invitation process for the existing user to the community with KYC enabled

### LINKED ISSUE #312: Change invitation flow for existing users 15

amiskov:
Closes #312 

If the inviter's community has KYC enabled then check if invitee (the user who is invited) has all KYC fields filled. If these fields are not filled then show the form with related fields and let the invitee join the community only if all the required fields are filled.

See the discussion in #312 for more details.

⚠️ The backend part is not fully implemented yet. To test the KYC form go to `src/elm/Page/Community/Invite.elm:626` and change `hasCommunityKycEnabled` to `True`. This branch is supposed to be merged since it doesn't change anything for existing users and finalized in the separate issue (TODO: add the number).

⚠️ Most of the translations are added by @RodrigoLeiteF in #311, so it's reasonable to add them afterward to avoid conflicts in translation files.

- Make sure you are logged in.
- Try to join the community with KYC enabled (see the ⚠️ note above) without having KYC fields filled.
- Also, try other variants:
    + Try to join the community where you are already a member to see the notice "Sorry, you are already a member";
    + Try to join the community without KYC enabled to see no changes in invitation process;
    + Try to join the community with KYC enabled having all KYC fields filled to see no form.


amiskov:
I marked this PR as ready for review and I suggest to do the following:

1. Merge this PR into `master` after the review so we could use modules for country numbers on the Profile and Register pages (at least for validation).
2. Create a new issue for implementing the communication with the backend part when it's ready.

There are modules in the `KYC/CostaRica` folder for DIMEX, NITE, Phone, and Cedula de Identidad. They may seem to be overkill for now, but I think they could be helpful for code reusing (validation and maybe for encoding/decoding). I suggest to leave them as-is for now, try to use them on the Profile and Register pages and see how it will work for us.

PR #341, opened 2020-08-20
New signup flow

### LINKED ISSUE #311: Update Sign Up process to use the new GraphQL mutation and new forms. 40

RodrigoLeiteF:
Closes #311


lucca65 (changes requested):
The image of the the side changes when I choose the user type, check the video:
Screen-Recording-2020-09-15-at-1


also:

- There is no mask, limitations to the length of characters to any fields in the form (like document, phone, etc)
- When I'm selecting a natural person, it asks for my company name
- Nothing happens when I fill in the form with all the required data
- All the form is cleared when I change between user types, even if a lot of the information is the same
- Its asking for address for the `natural` user type, when it shouldn't

src/elm/Page/Register.elm
```diff
@@ -683,64 +793,244 @@ update maybeInvitation msg model guest =
...
+                    { input = InputObject.buildKycDataUpdateInput obj }
+                    Graphql.SelectionSet.empty
+                )
+                CompletedKycUpsert
+    in
+    case formType of
+        Juridical form ->
+            cmd
+                { accountId = form.account
+                , countryId = "1"
```
amiskov, line 959: @RodrigoLeiteF  there will be a type mismatch after merging the latest changes from the `kyc` branch since the type for `county_id` and other address fields were changed to `:id` instead of `:string`.

I already had this mismatch somewhere, it can be fixed just by importing `import Cambiatus.Scalar exposing (Id(..))` and using `countryId = Id "1"`.

amiskov:
@RodrigoLeiteF regarding this comment above:

> The image of the the side changes when I choose the user type, check the video

I guess it could be fixed by having `h-screen` instead of `min-h-screen` in `src/elm/Session/Guest.elm:143`. The picture jumps because of the height of the content changes.

src/elm/Page/Register.elm
```diff
@@ -476,16 +522,40 @@ type alias UpdateResult =
...
     | AgreedToSave12Words Bool
     | DownloadPdf PdfData
     | PdfDownloaded
     | CopyToClipboard String
     | CopiedToClipboard
+    | CompletedLoadInvite (Result (Graphql.Http.Error (Maybe Invite)) (Maybe Invite))
+    | CompletedLoadCountry (Result (Graphql.Http.Error (Maybe Address.Country)) (Maybe Address.Country))
+    | AccountTypeSelected AccountType
+    | FormMsg EitherFormMsg
+    | DefaultFormMsg1 DefaultForm.Msg
```
amiskov, line 538: Looks like `DefaultFormMsg1` and `EitherFormMsg` are somehow similar in working with the Default form. Maybe we use `EitherFormMsg` only?

amiskov:
## Translations
image

Zip for `pt-BR`:
image

## Doc Numbers
NITE has 10 digits but on the first load of the page the counter shows 9:
image

Also, while changing the document type and entering numbers there may happen other mismatches in entered/required digits like this:
image
This state doesn't show an error message after pressing the "Continue" button.

lucca65 (changes requested):
It don't follow the design:

Screen Shot 2020-09-17 at 11 41 06


- [x] It should have the inviting community on the header
- [x] It should have the "steps" (step 1 of 2 / basic informations) when you haven't selected a type yet
- [x] Missing text "How would you like to register
- [x] In this version the "step" beneath the buttons, it would be above
- [x] There are spaces on the left side of the "Natural person" button and on the right side of "Juridical person"
- [x] There is a divider after the steps, which don't exist on the design
- [x] The spacing between the buttons is bigger than in the design
- [x] in the State select input the text is just "Select" which is not what the design specify
- [x] You can put letters in the document input
- [x] Phone number has no mask
- [x] When I don't fill in the district the validator shows validates it but no error is shown
- [x] When creating a juridical account, after I've received my twelve words, and clicked "Create my account" button, nothing happens

RodrigoLeiteF:
@amiskov For the invalid state in the document field, I made a change to reset the field to an empty state when the user changes the document type. Do you think that's a good enough fix?

amiskov:
> @amiskov For the invalid state in the document field, I made a change to reset the field to an empty state when the user changes the document type. Do you think that's a good enough fix?

@RodrigoLeiteF Yeap, from my perspective this behavior is much better!

src/elm/Session/Guest.elm
```diff
@@ -151,15 +202,38 @@ viewLeftCol mdWidth =
...
+    in
+    header
+        [ class "flex items-center justify-between pl-4 md:pl-6 py-3 bg-white" ]
+        [ a [ Route.href (Route.Login Nothing) ]
+            [ case model.community of
+                Loading ->
+                    loadingSpinner
+
+                _ ->
+                    imageElement logo
```
amiskov, line 236: I think we could be a bit trickier here.

Now only the default Cambiatus community is shown as [the design](https://www.figma.com/file/47BEsGEMSWfhTm3zC0ugyA/Ticolones-version?node-id=3%3A2206) suggests:
image

Since the Ticolones is an important community, we have to make sure that it looks according to the branding. We could have an ad-hoc `if` statement for the `TICO` symbol and show the SVG logo stored in the repo:
image

For all other communities, it would be great to show a title next to the logo:
image

I tried this code and it worked pretty fine:

```elm
Loaded c ->
    span [ class "whitespace-no-wrap" ]
        [ imageElement logo
        , span [ class "inline-block align-middle ml-1 font-bold capitalize" ]
            [ text c.title ]
        ]
```

RodrigoLeiteF: I talked to Helton about this on Slack, and we came to the conclusion that it's best to just show the community logo, as that lets community owner upload logos with their own fonts and styles if they wish to do so.

amiskov: Ok, that seems reasonable. But what about cases like this where we show only the sign (icon) without the text part?
image

amiskov (approved):
I think we can merge everything together to have all the pieces of the KYC feature in one branch. I'm sure we will find a bunch of bugs but I think it would be more convenient to continue our work having all the pieces of the puzzle combined. Regarding [my comment about community logos](https://github.com/cambiatus/frontend/pull/341#discussion_r491103888), we can figure out the solution later, in a separate issue.

src/elm/Session/Guest.elm
lucca65: maybe we could have two logos, with default sizings. But for now we may be inconsistent

lucca65:
+1 on @amiskov 's suggestion. Still al lot to fix and improve

lucca65 (approved):

src/elm/Session/Guest.elm
RodrigoLeiteF: According to him that's not necessarily an issue, and we should just let users know of what the logo should contain. @heltonlr can you explain it here please?

heltonlr: I said that we cannot use text for the logo, the correct is to use an image.  If the "ICON + TEXT"a logo gets bad due to the platform limitation, we can only use the "ICON"s logo (as we did with Muda).

PR #342, opened 2020-08-21
Form module

RodrigoLeiteF:
N/A

Adds a centralized API for creating forms in the Cambiatus style

N/A

### DIFF
--- a/src/elm/View/Form/Input.elm
+++ b/src/elm/View/Form/Input.elm
@@ -92,7 +92,11 @@ input options =
         []
 
 
-{-| Adds a character counter to your input. For more information, see the InputCounter module
+{-| Adds a character counter to your input. This does not limit the amount of characters automatically.
+Validation must be done in your onInput handler.
+
+For more information, see the InputCounter module
+
 -}
 withCounter : Int -> InputOptions a -> InputOptions a
 withCounter maximum options =
--- a/src/elm/View/Form/Input.elm
+++ b/src/elm/View/Form/Input.elm
@@ -7,8 +7,6 @@ import Session.Shared exposing (Translators)
 import View.Form
 
 
-{-| Required options for an input
--}
 type alias InputOptions a =
     { label : String
     , id : String
@@ -22,6 +20,8 @@ type alias InputOptions a =
     }
 
 
+{-| Required options for an input
+-}
 type alias RequiredInputOptions a =
     { label : String
     , id : String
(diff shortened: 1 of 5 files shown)


src/elm/View/Form/Input.elm
```diff
@@ -0,0 +1,97 @@
...
+                    [ text <|
+                        options.translators.tr
+                            "edit.input_counter"
+                            [ ( "current", String.fromInt <| String.length options.value )
+                            , ( "max", String.fromInt number )
+                            ]
+                    ]
+
+            Nothing ->
+                text ""
```
amiskov, line 68: I suggest moving the counter into a separate module (it could be used also for textareas) and use it like this:

```elm
InputCounter.view tr 12 form.username
```
The module itself:
```elm
-- src/elm/View/Form/InputCounter.elm
module View.Form.InputCounter exposing (..)

import Html exposing (Html, div, text)
import Html.Attributes exposing (class)
import I18Next


view : (String -> I18Next.Replacements -> String) -> Int -> String -> Html msg
view tr max str =
    div [ class "input-counter" ]
        [ text <|
            tr "edit.input_counter"
                [ ( "current", String.fromInt <| String.length str )
                , ( "max", String.fromInt max )
                ]
        ]
```

src/elm/View/Form/Input.elm
```diff
@@ -0,0 +1,97 @@
...
+        , ul []
+            (options.problems
+                |> Maybe.withDefault []
+                |> List.map viewFieldProblem
+            )
+        ]
+
+
+input : InputOptions a -> Html a
+input options =
```
amiskov, line 78: It doesn't look like input. It's batteries included form field for input of type `"text"`. Maybe it's better to rename it to `textField`?

amiskov (changes requested):
I think we can improve the `Form.View.Input` module beyond the given comments but I don't think we have time for this now :( It would be great to get back to the `Form` module later when we have more time! And for now, I think this PR could be merged after some minor fixes and improvements.

Also, examples of the actual initialization in the doc-comments for `input` and `select` modules could be very handy!

src/elm/View/Form/Input.elm
RodrigoLeiteF: I think it's best to leave it as `input` to keep it consistent with what you'd expect from an `<input>` HTML tag. A textField for example could be read as a text area.

amiskov (approved):
Looks good! I think we should try to use this module and see what may be added/improved while working with this way of creating elements.

PR #345, opened 2020-08-27
Feature/graphql signup

lucca65:
N/A

- Make the Frontend signUp process go through GraphQL instead of Rest. The new GraphQL API now properly works with invitations and regular sign ups.
-


amiskov (changes requested):
The code looks much cleaner!

It seems there could be a problem. I tried to register some users from this branch on my localhost and sometimes I was unable to login with the provided passphrase:
image
In this case, I've got two errors instead of one, as it was before:
image
This happens not each time. I have successfully registered several accounts as well. It looks like `eos.getAccout` method is called twice sometimes...

It will probably happen when the website is loaded for the 1st time. E.g. when I go to incognito mode and trying to register most likely I'll get the error. If I try to do the same for the 2nd time in the current tab (same incognito tab), then most likely I'll register successfully 🤷🏻‍♂️

How to reproduce:
- clear all the browser cache or go to incognito mode
- try to register with or without the invitation
- fill in the registration form, receive the passphrase and try to login with it
- if in the console were 2 errors, then you couldn't log in and will get `Account not found` error.

UPD: Some more info on this mysterious error:

- for some reason, sometimes account is generated twice;
- for the first time it generates an account with all the data from the registration form;
- sometimes it fires the generating process once again and generates keys and the passphrase for the _empty account_;
- the passphrase for the empty account is shown to the user but it doesn't work when the user tries to log in with it.

lucca65:
Great finding @amiskov this happened a few times with me, but I wasn't able to pin exactly what made it happen because it doesn't happen always (maybe due to the async nature of javascript engines). MUDA also had this problem a few times, so this is a very good finding.

amiskov (approved):
Looks good 👍

PR #346, opened 2020-08-28
Add KYC fields to the user profile pages

### LINKED ISSUE #310: Update Private Profile to display filled in KYC information 10
Display new KYC information if the community the user has selected has enabled it. 

https://www.figma.com/file/47BEsGEMSWfhTm3zC0ugyA/Main-flow?node-id=948%3A468

Update from the 3rd September, related to the Edit Profile page:

- All KYC data and the Name of the User _should not be editable_. We need to keep these fields disabled because they identify the user. User can fill Name and KYC fields only once: during the registration for new users, and during the invitation process to the KYC-enabled community [for the existing ones](https://github.com/cambiatus/frontend/issues/312).
- Users can’t edit KYC data, but they should be able to _remove KYC data_. Use [this design](https://www.figma.com/file/47BEsGEMSWfhTm3zC0ugyA/Ticolones-version?node-id=948%3A468) for the Profile page.
- There is a probability for the users to make mistakes in their names or phone numbers. Also, existing users could have nicknames instead of their real names. Since we are making these fields disabled we should provide some way to contact us to ask for help correcting these kinds of mistakes.
---
- It’s ok to have Costa-Rican fields in the user Profile for now since we have KYC only for the users from Costa-Rica. We will improve it when we will actually need it.
- We have `is_verified` field. We could use it later if we need it. E.g. allow users to edit their KYC fields and the Name if the profile is not verified yet.

amiskov:
Closes #310 

This PR handles only showing (not editing/deleting) KYC and Address information on the User Profile pages.

Changes:
- Show user's KYC info on the Private Profile page
- If the user has an Address, then show it on the Private and Public Profile pages
- There are also minor changes on the Edit Profile page, mostly code organization. These changes will be useful while working on the frontend part of cambiatus/backend/issues/60. This looks like a separate issue for the frontend, so I didn't implement it in this PR.

For now, there is no map on the Profile page, only an address as a string. This is intentional since implementing the map will take some time. We will back to the map after finishing more important tasks.
image

- Fill some KYC and/or Address data somehow (for now it's possible from the DB, `kyc_data` and `address` columns)
- Go to the user profile page, there should be KYC/Address information.
- Also, check the Public profile page. Public Profile should not show KYC data, but if the user has an Address, then it should be visible on the Public Profile pages.


src/elm/Page/Profile/Editor.elm
```diff
@@ -121,13 +121,26 @@ view_ loggedIn model profile =
...
+                hasCommunityWithKyc =
+                    profile.communities
+                        -- TODO: use .hasKyc instead of `True` when it's ready
+                        |> List.any (\c -> True)
+            in
+            if hasCommunityWithKyc then
+                viewDisabledInput
+
+            else
+                viewInput
```
lucca65, line 135: nice!

src/elm/Kyc.elm
```diff
@@ -0,0 +1,40 @@
...
+import Cambiatus.Object
+import Cambiatus.Object.KycData as KycData
+import Graphql.SelectionSet as SelectionSet exposing (SelectionSet, with)
+import Json.Decode as Decode exposing (Decoder, bool, string)
+import Json.Decode.Pipeline exposing (required)
+
+
+type alias ProfileKyc =
+    { userType : String
+    , documentType : String
```
lucca65, line 16: Those two could be types, but I'm 100% ok of keeping them a string for now

lucca65 (approved):

PR #350, opened 2020-09-11
Restrict access to KYC-enabled communities for users without KYC info filled

### LINKED ISSUE #313: Make the app know if the user don't have KYC filled in 15
In communities that have KYC enabled we need to find a way to block users actions if their profile is not filled in yet.

This can happen to existing users that removed their KYC information due to privacy reasons

amiskov:
Closes #313 

Lock community pages for users without KYC if the community has KYC enabled. Leave access only to static pages, Profile, and Edit Profile.

There should be no changes, if:
- The current community has KYC enabled and the user has KYC filled.
- The current community has KYC disabled.

There should be the lock screen, if:
- The current community has KYC enabled and the user hasn't his KYC filled.


lucca65 (approved):
Very good! It uses the existing structure very well! Congrats @amiskov for the elegant code!

PR #363, opened 2020-09-20
Parametrize contract call

lucca65:
Closes N/A

- [x] Parametrize contracts
- [x] Add has_kyc flag only to community creation (cannot be edited)
- [x] Improve a little bit the designs of the community creation form

### DIFF
--- a/src/elm/Community.elm
+++ b/src/elm/Community.elm
@@ -523,8 +523,7 @@ encodeCreateCommunityData c =
         , ( "invited_reward", Eos.encodeAsset c.invitedReward )
         , ( "has_objectives", Eos.encodeEosBool c.hasObjectives )
         , ( "has_shop", Eos.encodeEosBool c.hasShop )
-
-        -- , ( "has_kyc", Eos.encodeEosBool c.hasKyc )
+        , ( "has_kyc", Eos.encodeEosBool c.hasKyc )
         ]
 
 
--- a/src/elm/Community.elm
+++ b/src/elm/Community.elm
@@ -554,7 +554,8 @@ type alias UpdateCommunityData =
     , invitedReward : Eos.Asset
     , hasObjectives : Int
     , hasShop : Int
-    , hasKyc : Int
+
+    -- , hasKyc : Int
     }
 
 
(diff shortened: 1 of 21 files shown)


src/elm/Community.elm
```diff
@@ -569,7 +570,8 @@ encodeUpdateLogoData c =
         , ( "invited_reward", Eos.encodeAsset c.invitedReward )
         , ( "has_objectives", Encode.int c.hasObjectives )
         , ( "has_shop", Encode.int c.hasShop )
-        , ( "has_kyc", Encode.int c.hasKyc )
+
+        -- , ( "has_kyc", Encode.int c.hasKyc )
```
amiskov, line 574: I guess we don't need this line. Also, maybe it's better to rename `encodeUpdateLogoData` funciton to `encodeUpdateCommunityData`?

src/elm/Community.elm
```diff
@@ -554,7 +554,8 @@ type alias UpdateCommunityData =
     , invitedReward : Eos.Asset
     , hasObjectives : Int
     , hasShop : Int
-    , hasKyc : Int
+
+    -- , hasKyc : Int
```
amiskov, line 558: It looks like this line is obsolete.

src/index.js
```diff
@@ -67,6 +67,7 @@ const PUSH_PREF = 'bespiral.push.pref'
 const SELECTED_COMMUNITY_KEY = 'bespiral.selected_community'
 const env = process.env.NODE_ENV || 'development'
 const config = configuration[env]
+console.log(flags())
```
amiskov, line 70: I guess we don't need this debugging info anymore.

amiskov (changes requested):
Found a few small issues.

src/elm/Community.elm
lucca65: Actually at some point we will allow communities to enable KYC, so that's why I left it commented

src/elm/Community.elm
lucca65: The same with the previous comment

src/index.js
lucca65: Good catch!

src/elm/Community.elm
amiskov: Oh, I see!

amiskov (approved):

PR #366, opened 2020-09-22
Fix registration breadcrumb

RodrigoLeiteF:
Closes N/A

Changes the text for the step breadcrumb when user advances to step 2

1. Go to the registration page and create an account

### DIFF
--- a/src/elm/Page/Register.elm
+++ b/src/elm/Page/Register.elm
@@ -295,8 +295,19 @@ viewCreateAccount translators model =
 
                 _ ->
                     div [] []
+
+        backgroundColor =
+            case model.step of
+                1 ->
+                    "bg-white"
+
+                2 ->
+                    "bg-purple-500"
+
+                _ ->
+                    "bg-white"
     in
-    div [ class "flex flex-grow bg-white flex-col" ]
+    div [ class ("flex flex-grow flex-col " ++ backgroundColor) ]
         [ viewTitleForStep translators model.step
         , case model.status of
             LoadedAll invitation _ ->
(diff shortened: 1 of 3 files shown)


amiskov:
It looks like there are some issues with the design. Here's the 2nd step from `master` branch:
image
And here is the current branch:
image

RodrigoLeiteF:
Ah, I didn't realize we already had that in our current flow, and since it's not on Figma I assumed it would look the same for both steps. I'm on it

RodrigoLeiteF:
Submitted a fix that changes it to look like this:
image

According to Helton the placement is fine, since it stays consistent with the previous step on the flow.

amiskov:
> since it's not on Figma I assumed it would look the same for both steps. I'm on it

It was in Figma, I was implementing [this design](https://github.com/cambiatus/frontend/issues/151#issuecomment-614108671) some time ago, but it's lost now :(

src/elm/Page/Register.elm
```diff
@@ -295,8 +295,19 @@ viewCreateAccount translators model =
 
                 _ ->
                     div [] []
+
+        backgroundColor =
+            case model.step of
+                1 ->
+                    "bg-white"
```
amiskov, line 302: This case is already covered by `_`.

amiskov (changes requested):
All looks good, I have only one little suggestion.

src/elm/Page/Register.elm
RodrigoLeiteF: Fair enough :p

amiskov (approved):

PR #368, opened 2020-09-22
Only request country information on kyc-enabled communities

RodrigoLeiteF:
Closes N/A

Limits registration form to only request country information if invited community is KYC-enabled.

1. Open the browser network console
2. Go to a community with no kyc enabled
3. Look at the requests

### DIFF
--- a/src/elm/Page/Register.elm
+++ b/src/elm/Page/Register.elm
@@ -39,10 +39,7 @@ queries : Maybe String -> Shared -> Cmd Msg
 queries maybeInvitationId shared =
     case maybeInvitationId of
         Just invitation ->
-            Cmd.batch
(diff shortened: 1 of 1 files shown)


src/elm/Page/Register.elm
```diff
@@ -557,6 +554,11 @@ pdfData keys =
     }
 
 
+dummyCountry : Address.Country
+dummyCountry =
```
lucca65, line 558: I don't think this is a good idea, we can know from the invitation what is the community, and thus know if it have KYC enabled, we can do better!

lucca65 (changes requested):

amiskov (approved):

lucca65 (approved):

PR #386, opened 2020-10-06
Upgrade node to 14

### LINKED ISSUE #338: Bump NodeJS version
Check if its possible to use a newer version of NodeJS to build our project, last time we've checked a packaged required us to use node 12. A lot of security patches have been introduced

RodrigoLeiteF:
Docker doesn't provide an image for Node 14.8.0, so we remove the minor
version specification from package.json

We also remove the installation of yarn as `node:14` already comes with
it installed

----

Closes #338

1. Upgrades node version to 14

1. Run `docker build .`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 #================
 # Build Stage
 #================
-FROM node:12.0-alpine as builder
+FROM node:14-alpine as builder
 
 ENV NODE_ENV=production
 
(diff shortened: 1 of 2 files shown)


amiskov (commented):
It looks like we build assets for the new infra without the Docker image. Maybe we don't need Docker config at all... Anyway, this needs to be discussed, I will ask about it at the meeting.

lucca65:
As discussed on today's meeting we decided to completely move away from docker, so those files will be removed on #388

PR #396, opened 2020-10-13
Feature/remove docker

### LINKED ISSUE #388: Remove docker configs/files

lucca65:
Closes #388 

Removes docker references from the project, with the new infra they are no longer necessary.


amiskov (approved):

amiskov:
@lucca65 @victorgdev Since we are cleaning stuff and updating the docs I think it's a good time to make sure we all understand our frontend configs correctly :)

Now we have several sources for config data. Here's how I understand the purpose of each of them (@lucca65 please correct me if I'm wrong):

- `src/scripts/config.js` is the bottom level config. All the defaults are stored here for `DEV`, `PROD`, and `LOCAL` environments.
- `env-config.js` is the config for the dev/prod/demo servers. This file overwrites the defaults from the `config.js` above. Currently, this file _is not used_ in the local dev environment, on the localhost it won't be loaded correctly. In the repo, this file contains the data for the Staging server (in Demo it will be different, in Muda, it will be different, etc.).
- `src/elm/Flags.elm` is for... TBH I don't understand why we need to store some config details here (endpoints, logo path, etc.).

I still have some dark spots in understanding our configuration for the frontend. @lucca65 I'll much appreciate your help in clarifying this! Here are some questions:

- What is the difference between `local` and `dev` environment? I guess `local` is for the case when the developer has all the servers (eos, backend, even source) on the local machine, am I right?
- Do we really need to store config data in `src/elm/Flags.elm`? Maybe we can just use the data via port to eliminate the duplication?
- Maybe we shouldn't store `env-config.js` in the repo since it's different for each server/environment? I think we could have `env-config.js.sample` template in Git and create the real `env-config.js` on each server by hand from it.

lucca65:
@amiskov I'll add your clarification to our README.md along with the answer to the questions

PR #397, opened 2020-10-13
Feature/min balance community creation

lucca65:
Closes N/A. Depends on #396 

- Add ability to specify the community min balance during community creation

- Open `/community/new`
- Fill in all the fields, including the new min balance field
- Create community
- open [The block explorer](https://local.bloks.io/account/cambiatus.tk?loadContract=true&tab=Tables&table=stat&account=cambiatus.tk&scope=MIN&limit=100&nodeUrl=staging.cambiatus.io&coreSymbol=SYS&systemDomain=eosio) and check if the min balance is correct


src/elm/Page/Community/Editor.elm
```diff
@@ -140,6 +142,7 @@ editForm community =
     , logoList = logoList
     , inviterReward = community.inviterReward
     , invitedReward = community.invitedReward
+    , minBalance = community.minBalance |> Maybe.withDefault -100
```
amiskov, line 145: Here we have `-100` as the default value, but for the `newForm` we have `-10`.

Maybe we can do this?
```elm
, minBalance = community.minBalance |> Maybe.withDefault newForm.minBalance
```

src/elm/Page/Community/Settings/Features.elm
```diff
@@ -298,6 +298,7 @@ saveFeature feature state authorization { shared, accountName } community =
             , description = community.description
             , inviterReward = community.inviterReward
             , invitedReward = community.invitedReward
+            , minBalance = community.minBalance |> Maybe.withDefault -100
```
amiskov, line 301: Same here: `-100` is used as the default value but in the form for the new community `-10` is used.

amiskov (changes requested):
Spotted a couple of issues that are hardly possible but I think it worth adjusting them.

src/elm/Page/Community/Editor.elm
lucca65: Done, this is way better

src/elm/Page/Community/Settings/Features.elm
lucca65: I'll change the default on the form and keep it like this

amiskov (approved):

PR #398, opened 2020-10-14
Fix/juridical invitation

lucca65:
Looks like we did a bad merge at some point. Lost the ability to send invitation details.

Add user_type to sign up


amiskov (approved):
Looks good!

PR #400, opened 2020-10-14
Update README.md

lucca65:
@amiskov here is the answers, maybe we can add those to the README.md, let me know your thoughts

> What is the difference between `local` and `dev` environment? I guess `local` is for the case when the developer has all the servers (eos, backend, even source) on the local machine, am I right?

This actually existed because of the our old Docker env. Docker behaves a little different from the dev env, but know that we removed it, we no longer need the local abstraction

> Do we really need to store config data in `src/elm/Flags.elm`? Maybe we can just use the data via port to eliminate the duplication?

We can choose to fail the whole app in the case of a failed parsing the Flags, today, if we fail to parse, we just load the default ones. It was made like this to always have a working build even if the configs were mistyped. Thinking back this is kinda of a [Defensive programming](https://en.wikipedia.org/wiki/Defensive_programming) way of thinking, I and don't like it

> Maybe we shouldn't store `env-config.js` in the repo since it's different for each server/environment? I think we could have `env-config.js.sample` template in Git and create the real `env-config.js` on each server by hand from it.

Agreed


lucca65:
@victorgdev does the new instructions are clear enough? Can we improve?

src/scripts/config.js
```diff
@@ -1,35 +1,4 @@
...
-  logoMobile: '/images/logo-cambiatus-mobile.svg',
-  bespiralAccount: 'bespiral',
-  communityContract: 'bes.cmm',
-  tokenContract: 'bes.token',
-  allowCommunityCreation: true,
-  selectedCommunity: '0,BES',
-  pushKey:
-    'BDzXEdCCYafisu3jmYxBGDboAfwfIHYzM9BbT2DmL8VzIqSWu5BnW6lC-xEoXExQUS81vwOSPF9w8kpcINWCvUM'
-}
-
```
amiskov, line 32: Less possibility for misunderstanding 🎉

README.md
```diff
@@ -42,3 +42,12 @@ elm-analyse -p 3001 -s   # show extended analyse in the browser on `localhost:30
 yarn install
 yarn build
 ```
+
+## Configurations
+
+Our app have a somewhat complex configuration stack, allowing the app to run without anychanges, connecting to the staging environmnent. Here is an outline of all the configuration files and how the interact:
```
amiskov, line 48: `anychanges` → `any changes`

README.md
```diff
@@ -42,3 +42,12 @@ elm-analyse -p 3001 -s   # show extended analyse in the browser on `localhost:30
...
+
+- `src/scripts/config.js` is the bottom level config. All the defaults are stored here for `DEV`, `PROD`, and `LOCAL` environments.
```
amiskov, line 50: We don't need `local` anymore :)

README.md
```diff
@@ -42,3 +42,12 @@ elm-analyse -p 3001 -s   # show extended analyse in the browser on `localhost:30
...
+- `env-config.js` is the config for the dev/prod/demo servers. This file overwrites the defaults from the `config.js` above. Currently, this file _is not used_ in the local dev environment, on the localhost it won't be loaded correctly. In the repo, this file contains the data for the Staging server (in Demo it will be different, in Muda, it will be different, etc.).
```
amiskov, line 51: Since newcomers are hardly familiar with what Muda is, I think the sentence in parentheses could be like this: "_(in Demo server it will be different, in SomeCommunity server it will be different, etc.)_".

amiskov (changes requested):

src/scripts/config.js
lucca65: 🔥

amiskov (approved):
Nice!

PR #434, opened 2020-11-17
Improve Sentry integration

### LINKED ISSUE #427: Improve Sentry integration
We got a recurring situation during our support for our users: problems during login or something during the claim process that went wrong and we have no way of finding out what's wrong. Usually we use Sentry to check what went wrong, but we turn out to have tons of events there and its hard to find the exact event we are trying to debug, and even so, sometimes we got no context at all of what's wrong.
Its also very common to find situations where blockchain errors just merge together under the same General event, making it hard to know what's exactly going on 

-  [Send current logged user to Sentry](https://docs.sentry.io/platforms/javascript/enriching-events/identify-user/). It would be awesome to know which of the events we got are the one of the user we are supporting right now.
- Ungroup EOS errors So the blockchain message appears as the error on Sentry
- Add [breadcrumbs](https://docs.sentry.io/product/error-monitoring/breadcrumbs/) that we can add so we know the context of the action that failed
- Explore [Elm's](https://package.elm-lang.org/packages/romariolopezc/elm-sentry/latest/) Sentry package, maybe it can be easier to do some stuff

N/A. This is an exploratory issue 

**Additional context**
This discussion emerged on #419 and is part of it.

lucca65:
Closes #427 

- [x] Add env information
- [x] Add logged user
- [x] Add more details to EOS Transaction errors
- [x] Decouple EOS errors, now groups by blockchain response message
- [x] Add Elm internal errors (GraphQL, decode errors, etc)
- [x] Add **LOTS** of breadcrumbs so we can find our way home

- Break stuff!
- Check it on Sentry
- Evaluate if the info available is enough

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -119,7 +119,8 @@ function devLog (name, arg) {
 
 // Init Sentry
 Sentry.init({
-  dsn: 'https://535b151f7b8c48f8a7307b9bc83ebeba@sentry.io/1480468'
+  dsn: 'https://535b151f7b8c48f8a7307b9bc83ebeba@sentry.io/1480468',
+  environment: env
 })
 
 // Sentry Reporter
@@ -348,6 +349,7 @@ async function handleJavascriptPort (arg) {
       const passphrase = loginForm.passphrase
       const privateKey = ecc.seedPrivate(mnemonic.toSeedHex(passphrase))
 
+      // Storing stuff
       if (loginForm.usePin) {
         storePin(
           {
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -97,50 +97,64 @@ function flags () {
...
   if (env === 'development') {
     console.log('[dev]', name, arg)
   }
+
+  Sentry.addBreadcrumb({
+    message: typeof arg === 'string' || arg instanceof String ? arg : name,
+    level: 'info',
+    type: 'debug'
+  }
+  )
```
lucca65, line 133: I've changed this function that is used a lot in this file to be a general log. Sometimes the arg contain sensible data, so if its not a string, use the name.

src/index.js
```diff
@@ -97,50 +97,57 @@ function flags () {
...
   return flags_
 }
 
 // Start elm app with flags
 const app = Elm.Main.init({
   flags: flags()
 })
+Sentry.addBreadcrumb({
+  message: 'Started Elm app',
+  level: 'info',
```
amiskov, line 109: Since Sentry has `Severity` "enum" for the level I think it's better to use `Sentry.Severity.Info` and other "variants" all over the place. Now there are several `'info'` and several `Sentry.Severity.Info`.

```suggestion
  level: Sentry.Severity.Info,
```

amiskov (approved):
Nice! Do hope that it helps to clarify some things 🤞
There's only one minor suggestion about using Sentry "type" instead of JS string.

amiskov (approved):

PR #441, opened 2020-11-19
yet again I'm tired of javascript 

lucca65:
NA

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -156,7 +156,7 @@ app.ports.logError.subscribe((msg, err) => {
       scope.setExtra('Error shared by Elm', err)
       scope.setExtra('raw msg', msg)
       scope.setExtra('Parsed details', details)
-      Sentry.captureMessage(error)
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -143,11 +143,20 @@ app.ports.logError.subscribe((msg, err) => {
   if (env === 'development') {
     console.error(msg, err)
   } else {
+    let error = 'Generic Elm Error port msg'
+    let details = ''
+
+    if (Object.prototype.toString.call(msg) === '[object Array]') {
+      [error, details] = msg
+    }
+    Object.prototype.toString.call(msg)
```
amiskov, line 152: Looks like this string is redundant. Or maybe this expression should be assigned to a variable?

amiskov (changes requested):

src/index.js
lucca65: thanks for catching that!

amiskov (approved):

PR #452, opened 2020-11-25
Save selected community on login

### LINKED ISSUE #424: Fix inconsistent state on app load
Captura de Tela 2020-11-06 às 19 35 57

Not sure if its noticeable right away, but the token and the community don’t match. The community selector is also not available. Normal operations fail mysteriously.

instructions on how to replicate the bug:
First you have to get an invitation from any community other than Cambiatus. Just generate a new link, lets say for MIZU and logout.
Now open the invitation and register a new user. It should work as expected. Now reload the page. All is messy, no community selector, welcome message pointing to cambiatus, but your dear tokens, on the right place

amiskov:
Closes #424 by saving the current community to the `localStorage` for future use after the user is logged in. This will keep the current community selected after page reloading.

- [x] Show the correct logo of the community.
- [x] Show the Shop menu link and Objectives according to the community settings.

This could be achieved by performing the same actions as we do for selecting the community from the community selector. We should just trigger them while the user is logging in.

- Make sure your user is _a member of only one community_. The inconsistent state, described in #424 is related to that kind of user.
- Logout and then login.
- Reload the page. The community logo, shop, and objectives should be displayed correctly, according to the settings of the community.

### DIFF
--- a/src/elm/Session/LoggedIn.elm
+++ b/src/elm/Session/LoggedIn.elm
@@ -103,7 +103,8 @@ initLogin shared authModel profile_ =
     ( { model
         | profile = Loaded profile_
       }
-    , Api.Graphql.query shared (Community.settingsQuery selectedCommunity) CompletedLoadSettings
+    , Task.succeed selectedCommunity
+        |> Task.perform SelectCommunity
     )
 
 


lucca65 (approved):
😭 I love you bro

PR #457, opened 2020-12-01
Fix/long claim display

lucca65:
N/A

- Truncate text if its too long on the view claim
- Overflow hidden on long claim text on voting cards (/dashboard)
- Also changed the symbol display on the feedback message, it was including the symbol precision

## How to test ( a list of instructions on how to test this PR)
- Create an action with a very long description, but make sure it have long words (like one of those microsoft URLs, boy they sure love incredibly hard to share URLs)
- Claim it and make sure you are one of the validators
- See the error


lucca65:
also, the new Tailwind documentation is AWESOME

amiskov (dismissed):
The Dashboard looks good! I also noticed that there's a small issue on the Claim View page with those long words:
image

Maybe this PR is a good place to tweak the left margin for the header on the Claim View page as well?

lucca65:
+1! Added the fix for this case as well! thanks for noticing that @amiskov

amiskov (approved):
👍

PR #458, opened 2020-12-02
Fix duplicated claims on the Dashboard after logging in

### LINKED ISSUE #456: Duplicate claims after login
**Describe the bug**
We got duplicate claims on the first load of the dashboard.

**To Reproduce**
1. Make sure you have an account on cambiatus and that account is a validator on a single, open claim
2. Login 
3. You should see the claim duplicate
4. Reload the page in any other way and you will see only one

**Expected behavior**
We should see only one claim if the user has only one claim to validate. Although the *To Reproduce* steps say a single open claim it happens with any number of claims waiting for voting. The important part is that it only happens after the login

**Screenshots**
N/A

Put the URL of the page you are trying to use

All browsers and OS.

All devices

**Additional context**
I saw it happening during a debug session with ECOINS community. It was probably introduced by #424

amiskov:
Closes #456.

Redirecting to the Dashboard causes reloading all related data. We don't need it when the user logs in. We need it only when the community is changed by the top-left selector. This PR implements this behavior by adding a custom command to the `SelectCommunity` message.

See the steps to reproduce from #456. The claim cards on the Dashboard shouldn't be duplicated.

### DIFF
--- a/src/elm/Session/LoggedIn.elm
+++ b/src/elm/Session/LoggedIn.elm
@@ -103,8 +103,9 @@ initLogin shared authModel profile_ =
     ( { model
         | profile = Loaded profile_
       }
-    , Task.succeed selectedCommunity
-        |> Task.perform SelectCommunity
+    , Task.perform
+        (\_ -> SelectCommunity selectedCommunity Cmd.none)
+        (Task.succeed ())
     )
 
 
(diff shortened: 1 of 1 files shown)


lucca65 (approved):
🔥

PR #460, opened 2020-12-03
Remove old legacy_main.css

lucca65:
N/A

Remove Old css file, importing only what we still use and adjusting some stuff using tailwind classes

- Run the app
- Go through all pages to make sure everything is still looking as it should


amiskov (changes requested):
Found a couple of issues.

Highlighting of the dropdown item disappeared. The screenshot from Staging (in this branch we don't have blue highlighting):
image

---

On the `community/new` page there are problems with the icons for the communities. Default icons disappeared and the upload icon became gigantic:
image

Same problem with icons on the `community/<some-community>/edit` page (I guess they may be fixed at once).

Here's how it looks on Staging:
image

lucca65:
@amiskov I've transported the elements for the create community form (I didn't refactored it in tailwind because we will have to rework that screen at some point). I've also updated the style for the user selector to match the one we have on Figma (its actually pretty old, but haven't been updated on our code yet). Please let me know if you find anything else!

amiskov (approved):

PR #462, opened 2020-12-08
Update API usage on the app

lucca65:
N/A

This PR is a companion to the [backend counterpart](https://github.com/cambiatus/backend/pull/119), it updates the shop graphql objects. It also does one important renaming. Now `sales` is called `products` and `sale_history` is called `order`. This change isn't just semantic, in the next months we will work on improving the shop to allow orders to have multiple payment options as well as a statuses. So this change is the first step to do it, cleaning up the api to help future change. 

It also changes the scope of the shop. Now `products` query always require the `communityId` following up the change of having the app scoping everything to one community at the time. It also changes the query to show the users sales on the main view (we used to hide it, now its displayed alongside with all other products offers).

Another change is the removal of the thumbs from the shop listings. 

Just use the app normally, including shop parts. All should work as it should thanks to the strong elm type system

### DIFF
--- a/src/elm/Cambiatus/InputObject.elm
+++ b/src/elm/Cambiatus/InputObject.elm
@@ -456,6 +456,31 @@ encodeObjectiveInput input =
         [ ( "id", Encode.int input.id |> Just ) ]
 
 
+buildProductsFilterInput :
+    ProductsFilterInputRequiredFields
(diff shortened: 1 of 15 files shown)


amiskov (approved):
Looks good!

PR #477, opened 2021-01-25
Fix/staging tests

lucca65:
N/A Findings by @heltonlr during tests of our latest version

- Update GraphQL bindings
- Improve responsiveness for the "my products" link on dashboard
- Add missing translations on CAT an PT
- Fixed a problem where non validator would receive the voting buttons on Profile/claims
- Added page reload after enabling/disabling the Shop or Actions


amiskov (approved):
Looks good!
Nice catch on using `.1`, `.2`, etc. in translations! Noted to self.

PR #485, opened 2021-02-18
Feature/auth 316

### LINKED ISSUE #316: Add GraphQL Authentication
We need to update the GraphQL bindings on the frontend to accommodate the new Authentication method introduced by https://github.com/cambiatus/backend/pull/126

- [x] Update bindings
- [x] Contextualize queries with the logged user
- [x] Update sign_in/sign_up to grab the token and store it on the global state
- [x] Use GraphQL for sign_in instead of REST

N/A

**Additional context**
During the surge of users on Verdes, Authentication really became more and more important, this issue is part of the efforts to address this

amiskov:
Closes #316 

- Update GraphQL bindings to be consistent with [the backend](https://github.com/cambiatus/backend/pull/126/).
- Send the ad-hoc password with the singIn/signUp mutations and get/store the auth token from the backend.

- Make sure that your backend has the changes from the related [feature/316-add-auth](https://github.com/cambiatus/backend/pull/126/) branch. E.g. you can run backend locally and tweak `frontend/src/scripts/config.js:15` a bit.
- Try to sign in and sign up.
- Click around and make sure that links are working correctly.
- Try to log out and then login again.


src/elm/Page/Dashboard/Analysis.elm
```diff
@@ -593,12 +593,7 @@ fetchAnalysis { accountName, selectedCommunity, shared } { profile, statusFilter
                     Present filterRecord
 
         args =
-            { input =
-                { symbol = Eos.symbolToString selectedCommunity
-                , account = Eos.nameToString accountName
-                , filter = filter
-                }
-            }
+            { communityId = Eos.symbolToString selectedCommunity }
```
lucca65, line 596: filter here keep existing, right?

amiskov: Yeah, the filter will be moved into optional arguments according to the new types. I was pretty rough at updating GraphQl bindings, for now, wanted to satisfy the compiler first. I still need to revisit these changes, now it's partially working. Just wanted to have a draft PR to share. I like the changes so far! Less code, more straightforward and secure flow. Especially signUp/signIn.

config/env.js
```diff
@@ -72,7 +72,9 @@ function getClientEnvironment (publicUrl) {
         // For example, <img src={process.env.PUBLIC_URL + '/img/logo.png'} />.
         // This should only be used as an escape hatch. Normally you would put
         // images into the `src` and `import` them in code to get their paths.
-        PUBLIC_URL: publicUrl
```
lucca65, line 75: Usually I don't do this, I just edit `Flags.elm`,  `src/scripts/config.js` and `env-config.js.sample`

package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "build": "yarn bundle",
     "bundle": "NODE_ENV=production yarn webpack --color --mode=production --config config/webpack.config.prod.js",
-    "start": "GRAPHQL_SECRET=d8Ed.-qfhj7 standard && node scripts/start.js",
```
lucca65: this seems wrong

src/elm/Api/Graphql.elm
```diff
@@ -7,15 +7,28 @@ import Graphql.SelectionSet exposing (SelectionSet)
...
+            Graphql.Http.withHeader "authorization"
+                ("Bearer " ++ t)
+
+        Nothing ->
+            identity
+
+
 query : Shared -> SelectionSet a RootQuery -> (Result (Graphql.Http.Error a) a -> msg) -> Cmd msg
-query { endpoints } query_ toMsg =
+query { endpoints, authToken } query_ toMsg =
```
lucca65, line 22: maybe we can have two functions here: 

`publicQuery` and `query`. Because some queries won't have a token (like community query used for invitation) and others (the majority for sure) will have a token.

maybe:

```elm
query : LoggedIn.Model -> SelectionSet a RottQuery -> (. . .) -> Cmd msg
query { profile, shared } query_ toMsg =

publicQuery : Shared -> SelectionSet a  . . .
publicQuery {endpoints} query_ toMsg = 

```

This way the type signature already indicates us that some queries must be logged, while others don't.

Having the token saved to shared looks weird because we only got the token after logging in

src/index.js
```diff
@@ -83,13 +85,15 @@ function getUserLanguage () {
...
-  var flags_ = {
+  return {
     env: env,
+    graphqlSecret: graphqlSecret,
     endpoints: config.endpoints,
     language: getUserLanguage(),
     accountName: (user && user.accountName) || null,
     isPinAvailable: !!(user && user.encryptedKey),
     authPreference: window.localStorage.getItem(AUTH_PREF_KEY),
+    authToken: window.localStorage.getItem(AUTH_TOKEN),
```
lucca65, line 96: I'm a little bit confused. This `AUTH_TOKEN` is the token we get after calling `signUp` or `signIn`, or the password we send to those params?

amiskov: The [whole object](https://github.com/cambiatus/frontend/pull/485/files/e5eef2bbf8cadc930d6b140504fe1520c9975f5b#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R86) returned from `flags()` contains two properties related to authentication: `graphqlSecret` and `authToken`.

`graphqlSecret` is named after the config parameter [on the backend](https://github.com/cambiatus/backend/blob/44b7a3f8d9ac36bf50c467e051b574a5228f3702/config/dev.exs#L37
) and the `authToken` represents the response from the server with the token, generated by `CambiatusWeb.AuthToken.sign(user)` and could be stored/retrieved from the `localStorage`.

`graphqlSecred` is retrieved from the `GRAPHQL_SECRET` ENV variable during the build process.

src/elm/Api/Graphql.elm
amiskov: >maybe we can have two functions here

I tried that approach and I didn't like it. IMO the ideal signature would be:

```elm
query : Endpoint -> Maybe AuthToken -> SelectionSet a RootQuery -> ...
mutation : Endpoint -> Maybe AuthToken -> SelectionSet a RootQuery -> ...
```
`Maybe`s in the signature tell us everything we need to know and we don't have to support 2 functions with the same purpose. I tried a similar approach as well and I didn't like the increased complexity and the need to change the code in many places.

Also, we can't use `LoggedIn.Model` in `Api.Graphql` due to circular dependency, but we can achieve the same behavior with extensible records:

```elm
query : { loggedIn | shared : Shared, authToken : String } -> SelectionSet a RootQuery -> ...
```

I also tried it and didn't like it much... Although, it's nice not to have any logged-in related things in `Shared`, but we already have `Shared` "broken": there's `maybeAccount : (account, isPinAvailable)` in it. As an option, we can add the `authToken` as the third element to this tuple. I also tried it and I also didn't like it.

IMO, it's good enough to have `authToken` inside `Shared`. It allows us to keep all existing queries and mutations untouched: fewer changes, fewer errors.

package.json
amiskov: Yep, the variable should go before `node`:

```bash
standard && GRAPHQL_SECRET=d8Ed.-qfhj7 node scripts/start.js
```

This secret is just for the dev environment and it's the same as in the backend config. For the build we need to set this variable manually:

```fish
$ set GRAPHQL_SECRET=...
$ yarn build
```

config/env.js
amiskov: If we store the secret key that way then it will be very easy to find it in the source code. I introduced the `GRAPHQL_SECRET` env variable to be able to add it manually during the build process.

src/elm/Api/Graphql.elm
lucca65: +1 on the new signature suggested, but -1 on having more logged-in related things because we already have it. Also that isPinAvailable is reminiscent from a time where our app supported only logging in when we needed something signed, very different from the approach we have now. I would dare to say this isPinAvailable always have the same value.

Maybe we can get into a middle ground here, keep your suggestion of ignoring `LoggedIn`/`Share` altogether and instead use `Endpoint` and `Maybe`, then send the `Just token` when we have it available on the caller method

lucca65: if we want to avoid keep dragging this issue longer because it would require bigger changes than we can ask for @NeoVier 's help here, this refactoring is quite big

amiskov: @lucca65 Yep, these changes will definitely take some time. If I could pass this task along and do more on Natus, I'll be really thankful 🙏

NeoVier: I'm happy to help! Can I just get some context on what's left to do though?

lucca65: sure! This issue is the frontend implementation for https://github.com/cambiatus/backend/pull/126, that brings a basic API authentication flow to the app. Before this PR our API was public, but now certain queries will be scoped by the logged user and some mutations will only work if the user is changing its own data.

For example: we have an mutation that allows the user to delete their KYC data, before it would take one param `account` to know where it should delete the KYC data from, but now it has no arguments, taking instead the logged user. Another query is the `notifications` query, that took a `account` param but now has none.

The authentication flow is straight forward, whenever you call a `signUp` or `signIn` you will get a `Session` GraphQL object:

Captura de Tela 2021-03-02 às 08 14 10
Captura de Tela 2021-03-02 às 08 14 49

Session contains a `token` that must be stored inside LoggedIn (or not, as you can see in this thread here). Only loggedIn users will have a `token`.

We have queries that don't require authentication such as the the `invite` query, you can see which needs it by checking the schema documentation on the GraphiQL interface (I'm pushing a change today that includes that detail)

The queries look the same overall, usually only removing the `account` param or something similar and a new header is added to the request: `Authentication: Bearear #{token}`

We need to finish implementing this as @amiskov already did a lot of those changes and decide how we are going to support this new token we need to add to certain queries. Might be a good opportunity to change to the RemoteData pattern if it doesn't take a lot of time

let me know if you need a quick call or something similar too!

NeoVier: Alright, I'll take a look, and if I need more details I'll let you know!

NeoVier:
@lucca65 @amiskov Does this make more sense? I tried to move `authToken` into `LoggedIn`, which I think makes more sense (I still kept it in `Shared` for now, so I don't need to change every query yet), and sketched out the `query` function according to @amiskov (I think that's probably the best signature).

I just wanted to make sure this is a good way to go, since any other refactoring would be pretty large.

src/elm/Api/Graphql.elm
```diff
@@ -18,6 +19,14 @@ withAuthToken authToken =
             identity
 
 
+query2 : String -> Maybe String -> SelectionSet a RootQuery -> (RemoteData (Graphql.Http.Error a) a -> msg) -> Cmd msg
+query2 endpoint authToken query_ toMsg =
```
lucca65, line 23: maybe I'm a little bit decontextualized here but since you added the auth token to loggedIn, maybe we can just keep sending `Shared` to access `endpoints`?

src/elm/Auth.elm
```diff
@@ -825,9 +818,9 @@ update msg shared model =
         GotPinLogin (Ok ( accountName, privateKey )) ->
             UR.init model
                 |> UR.addCmd
-                    (Api.Graphql.query shared
-                        (Profile.query accountName)
-                        (CompletedLoadProfile (LoggedInWithPrivateKey privateKey) accountName)
+                    (Api.Graphql.mutation shared
+                        (signIn accountName shared)
+                        (CompletedSignIn (LoggedInWithPrivateKey privateKey))
```
lucca65, line 823: nice!

src/elm/Api/Graphql.elm
NeoVier: Yes, we definitely can! I don't think there are any immediate tradeoffs between one or the other, just that sending `Shared` might mean a tiny little less refactoring while transitioning from `query` to `query2`, so I guess that's better :sweat_smile:

lucca65 (approved):
Looks good awesome! Good to see `RemoteData` everywhere!! I'm loving the progress we are making in the app, this one is an old dream of mine, since the first versions of the app we talk about having this 🥲

@heltonlr we are going to need one of those big, complete tests here. This PR changes the permission and authentication methods so it affects every single page of the app, we need to make sure the user can only see/change their own data

NeoVier:
@heltonlr can we get another round of tests for this?

heltonlr:
> @heltonlr can we get another round of tests for this?

Sure @NeoVier , Which scenario would you like to be tested?

NeoVier:
Besides the issues you've pointed out last time, the payment history page (https://staging.cambiatus.io/payments/user, which is now [only accessible to logged in users](https://cambiatus.slack.com/archives/CA83HJAAD/p1615398335035700?thread_ts=1614954681.151200&cid=CA83HJAAD)), and also signing into a new community via invite link

lucca65 (approved):

PR #494, opened 2021-03-19
Feature/support subdomains

### LINKED ISSUE #470: Improve size of our loading Gif
This is not a bug but rather a optimization to improve our load time. It would be great to make the loading gif smaller, today it takes 502 kb

An animated SVG

Maybe a video? idk we should research

### LINKED ISSUE #482: Add community selector as a "welcome message"
Users sometimes don't know on which community they are in, adding a selector after the login would help a lot!

Make our community selector to be the first thing users see when login in on a community

NeoVier:
Addresses #495, closes #470

- [x] Load appropriate community based on URL
- [x] Store all community data only on `LoggedIn`
- [x] Remove symbols (like `0,CMB`) from URLs
- [x] Add new community configuration pages
  - [x] General information (name, description, subdomain, etc)
  - [x] Currency (inviter reward, invited reward, minimum balance)
  - [x] Refactor old edit page so it can only create new communities, and not edit existing ones
- [x] Auto redirect the user when changing communities
- [x] Share local storage with all domains
- [x] Improve loading gif
- [x] Change what happens when the user accesses the application:
  - [x] Logged in:
      - [x] User is member of the community: behave as usual
      - [x] User is not a member of the community, and community doesn't have autoinvites: show community selector (closes #482)
      - [x] User is not a member of the community, and community has autoinvites: show "Join community" page and let them sign in
  - [x] Guest:
      - [x] Community allows for autoinvites: show "Join community" page and let them sign in
      - [x] Community doesn't allow for autoinvites: redirect to https://www.cambiatus.com/

- Try accessing different communities by prepending the name of the community to the url (e.g. `muda.localhost:3000`) and seeing if the correct community is loaded. Try this with communities that:
    - You are a member of (should load the community fine)
    - You are not a member of and has auto invite (should invite you to join the community)
    - You are not a member of and doesn't have auto invite (should show the community selector)
- Try changing something (e.g. language) in one community, change to another community and see if the changes persist
- Try changing the community via the community selector, and see if you're redirected to the correct url
- Try creating a new community (via `/community/new`), and see if it behaves as expected
- Try changing the settings for your community and see if everything behaves as expected
- Without being logged in, try to access communities that have and don't have autoinvites - they should show the "Join community" page, and allow you to sign in if they allow for autoinvites
- Being logged in, try to access a community you're not a member of yet, but that allows for autoinvites - should show the "Join community" page, with the community information, and let you sign in to the community


public/translations/cat-CAT.json
```diff
@@ -886,4 +887,4 @@
     "no_transfers_found": "No s'ha trobat transferència segons els vostres criteris.",
     "pick_date": "Escolliu una data..."
   }
-}
```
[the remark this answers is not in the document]
NeoVier: I did configure this in my editor, and it shows up there :thinking: not sure why it gets deleted when I push. Any ideas?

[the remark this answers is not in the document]
lucca65: not sure maybe this can help! https://stackoverflow.com/questions/44704968/visual-studio-code-insert-new-line-at-the-end-of-files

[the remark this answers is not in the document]
NeoVier: Yeah, I ran into that and did that. Apparently it doesn't show up as a change, but the files now show up with a line break at the end :man_shrugging:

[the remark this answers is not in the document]
lucca65: you can try to use GitHub online editor and just pull from your machine then

NeoVier:
@lucca65 as this is a pretty big pull request, and to have better organization, I've created a feature branch for this, and I'll develop the other features needed for this in other PRs, let me know what you think

src/elm/Community.elm
```diff
@@ -405,12 +406,29 @@ encodeCreateCommunityData c =
         , ( "description", Encode.string c.description )
         , ( "inviter_reward", Eos.encodeAsset c.inviterReward )
         , ( "invited_reward", Eos.encodeAsset c.invitedReward )
+        , ( "min_balance", Eos.encodeAsset c.minBalance )
```
lucca65, line 409: yeah, this will not work bro, you see there are two calls related to the creation of a community: 

1. Creating a community with details like logo, name, and invitation stuff 
2. Creating a Cambiatus Token, that communities may use.

A Community is an abstraction that allows for some stuff like: creating sales, having a network of people, configure that network invitation rules, create objetives, validate claims etc.

A token is monetary abstraction, it includes: defining negative limits, configuring supply, issuing new tokens etc.

This call you are doing is related to [this action](https://github.com/cambiatus/contracts/blob/1931f370299bb3bda181bb45c979509991cb52bc/community/community.hpp#L165-L167), and as you can see there is no `min_balance`, and that's because this is not a community rule, but rather a token rule. 

to create a token you can [call this action](https://github.com/cambiatus/contracts/blob/1931f370299bb3bda181bb45c979509991cb52bc/token/token.hpp#L50) and this one has the min balance. 

this is good! this means you are starting to interact with the blockchain and there a few ways to discover how it works. You can use the [Bloks block explorer](https://local.bloks.io/account/cambiatus.cm?nodeUrl=http%3A%2F%2Fstaging.cambiatus.io&coreSymbol=SYS&systemDomain=eosio&loadContract=true&tab=Actions&account=cambiatus.cm&scope=cambiatus.cm&limit=100&action=create) to explore what the contract can do and how you can call the actions or query its tables. 

we use two contracts on Cambiatus: `cambiatus.cm` for the community stuff and `cambiatus.tk` for the token stuff

src/elm/Community.elm
```diff
@@ -405,12 +406,29 @@ encodeCreateCommunityData c =
...
+        |> required "name" Decode.string
+        |> required "description" Decode.string
+        |> required "inviter_reward" Eos.decodeAsset
+        |> required "invited_reward" Eos.decodeAsset
+        |> required "min_balance" Eos.decodeAsset
+        |> required "has_objectives" Eos.eosBoolDecoder
+        |> required "has_shop" Eos.eosBoolDecoder
+        |> required "has_kyc" Eos.eosBoolDecoder
+
+
```
lucca65, line 431: Why do we need a decoder for this?

lucca65 (changes requested):
I love the umbrella strategy bro! nice!

Just a single preliminary comment, since I wasn't able to test it locally. Is there anything I need to change on my hosts configuration or something similar? the app gets redirected to `app.localhost:3000` and my browser is unable to load anything

NeoVier:
> Is there anything I need to change on my hosts configuration or something similar? the app gets redirected to app.localhost:3000 and my browser is unable to load anything

It works out of the box for me. It might be something related to the browser you're using (it works for me on Firefox and Chromium), but you can try [editing your hosts file](https://stackoverflow.com/a/19016600). What kind of errors are you getting? Is it a browser thing or an app thing? If it's from the app, is there something on the console?

The only thing I haven't figured out is accessing the app from another device (I used to do this with the IP address `yarn start` gives you) - and I guess it's at least [not straightforward](https://stackoverflow.com/a/21499542)

lucca65:
looks that for some browsers it will not work at all. Do you think its possible to have a flag to disable it somehow? I mean, I can use chrome, but without some mechanism to disable it, how can we test stuff that is safari only? Somehow safari is the new IE 🤦 

Also, I've noticed that the redirect is happening using the community name, and it fails on certain communities that have spaces or special characters on them like `'` or `水`. Are you needing some type of modification on the backend right?

src/elm/Action.elm
```diff
@@ -828,24 +828,19 @@ generateVerificationCode actionId makerAccountUint64 proofTimeSeconds =
         |> String.slice 0 8
 
 
-isPastDeadline : Action -> Maybe Posix -> Bool
-isPastDeadline action maybeNow =
```
lucca65, line 832: much better to always have a date available

src/elm/Page/Community/Objectives.elm
```diff
@@ -99,12 +78,24 @@ view ({ shared } as loggedIn) model =
...
+                    Page.fullPageGraphQLError (t "community.objectives.title_plural") e
+
+                ( RemoteData.Loading, _ ) ->
+                    Page.fullPageLoading shared
+
+                ( RemoteData.NotAsked, _ ) ->
+                    Page.fullPageLoading shared
+
+                ( _, Loading ) ->
+                    Page.fullPageLoading shared
```
lucca65, line 98: nice!

src/elm/Page/Community/Settings/Currency.elm
```diff
@@ -0,0 +1,560 @@
...
+              }
+            , { accountName = loggedIn.shared.contracts.token
+              , name = "update"
+              , authorization = authorization
+              , data =
+                    { maxSupply = asset 21000000.0
+                    , minBalance = asset validModel.minimumBalance
+                    }
+                        |> Community.encodeUpdateTokenData
+              }
```
lucca65, line 219: this is the only call needed, the other one, on the `shared.contracts.community` is unnecessary on this screen

src/elm/Page/Community/Settings/Currency.elm
```diff
@@ -0,0 +1,560 @@
...
+                        }
+                        |> Input.withAttrs [ class "w-full" ]
+                        |> Input.toHtml
+                    ]
+                ]
+            , div [ class "bg-gray-100 py-4 text-center mb-10" ]
+                [ div [ class "text-xl font-medium space-x-4 mb-4" ]
+                    [ span [ class "text-black" ]
+                        [ text
+                            (String.fromFloat pi
```
lucca65, line 420: I like `pi` but I've found it to be confusing. @heltonlr do you think we can improve this?
Captura de Tela 2021-04-19 às 15 09 57


I mean I know that we can have more decimal places, but it might look confusing to see this: 3 MIZU your currency format

src/elm/Page/Community/Settings/Features.elm
```diff
@@ -323,6 +298,16 @@ saveFeature feature state authorization { shared, accountName } community =
     }
 
 
+receiveBroadcast : LoggedIn.BroadcastMsg -> Maybe Msg
+receiveBroadcast broadcastMsg =
+    case broadcastMsg of
+        LoggedIn.CommunityLoaded community ->
+            Just (CompletedLoadCommunity community)
```
lucca65, line 305: 🍬

src/elm/Community.elm
NeoVier: To create a new community, we use the `newCommunity` GraphQL subscription. So, if you look at [Community.Editor.Msg](https://github.com/cambiatus/frontend/pull/494/files#diff-4817c965f006ef54346aab990d3c1154f0573d7fb700e2ad7661c9c9e317ff84R574-R577), we fire a `Msg` when the user has `SubmittedForm`, which parses the model into a `CreateCommunityData` and, if it's valid, subscribes to the GraphQL `newCommunity` subscription, which then fires the `StartedCreatingCommunity` Msg, which takes that `CreateCommunityData`. To pass in `CreateCommunityData` to `StartedCreatingCommunity`, I'm using the port's `responseData` field, so we need to encode the `CreateCommunityData` when adding the port, and decode it on `jsAddressToMsg` to pass it to `StartedCreatingCommunity`.

I see two approaches we can take on this scenario:
- Do as I described above. This is nice because we don't have the impossible state where the user has submitted the form, and we started creating the community, but the form data isn't valid (e.g. `invitedReward` isn't a valid `Float`). The bad thing is we need to encode and then decode the data (which is why we need `min_balance` when encoding, and why we need this decoder). 
- Use the `Model` data. This is nice because we don't have to encode and then decode `CreateCommunityData`, but we'd have to parse the form data again, which leads to impossible states.

I still prefer passing data through `responseData` in the port, so we could pass in an object like so

```js
{
  createCommunityData : { /* Regular CreateCommunityData, as it was before this PR, without `min_balance` */ },
  minBalance: 0
}
```

That would mean we can keep the encoder without `min_balance`, and avoid those impossible states

NeoVier:
> looks that for some browsers it will not work at all. Do you think its possible to have a flag to disable it somehow? I mean, I can use chrome, but without some mechanism to disable it, how can we test stuff that is safari only? Somehow safari is the new IE :facepalm: 

I guess we could have a flag passed in on `yarn start` (or an env variable?). We'd need to store that in `Shared`, and then instead of redirecting when the user selects a community, we just load the correct one.

> Also, I've noticed that the redirect is happening using the community name, and it fails on certain communities that have spaces or special characters on them like ' or 水. Are you needing some type of modification on the backend right?

Yep, right now I'm just using the community's name, I'm just waiting until subdomain is a thing in the backend

src/elm/Community.elm
lucca65: We do use `newCommunity` subscription, yes, but that's only to get a callback from the server that event-source was able to capture the transaction.

The steps which we take:

1. Grab data from the user, validate form
2. Create a subscription before we actually save anything
3. Push the transaction to the blockchain with two actions (one creates the community and the other the token)
4. Wait for the event-source to capture it
5. Backend has a trigger configured, every time something is inserted to the `community` table it looks for subscriptions and publish them
6. The frontend grabs the event and redirects the user to the newly created community.

Those steps were created to ensure the user would be redirected to the community page after creating them, due to the async nature of the blockchain, if we redirect the user right away sometimes it may happen that they see a 404. This is not ideal but it means the GraphQL subscription is only something related to the automatic redirect, not the saving itself. The point is: to create community you don't need the `min_balance`. You will need it to create a token. 

This is why I find it wierd, that behavior haven't changed even after you added new forms and pages, so no need to have a new decoder to reach the same result we had before

lucca65:
> I guess we could have a flag passed in on yarn start (or an env variable?). We'd need to store that in Shared, and then instead of redirecting when the user selects a community, we just load the correct one.

Ok, seems reasonable!

> Yep, right now I'm just using the community's name, I'm just waiting until subdomain is a thing in the backend

@venomnert do you need any help with this?

src/elm/Page/Community/Settings/Currency.elm
NeoVier: What about updating `inviterReward` and `invitedReward` though?

src/elm/Community.elm
NeoVier: The old flow for creating a community was like this:

1. User sees empty form
2. User fills in form
3. User clicks "Save"
4. If the symbol field is actually a valid symbol, we subscribe to GraphQL's `newCommunity` (otherwise, literally do nothing)
5. We wait until the subscription port sends us something back.
6. a. If it sends "responded", just redirect to the community page (and the flow is over)
    b. If it sends "starting", we ask the user for authentication
7. Only now we validate the form
8. a. If the form is not valid, we show the errors to the user (and the flow starts from 2.)
    b. If the form is valid, we send the Eos transaction
9. a. If the Eos response was valid, redirect to community (we should remove this, so we only redirect when the subscription tells us to)
    b. Otherwise, show errors and start from 2.
    
The new flow is like this:

1. User sees empty form
2. User fills in form
3. User clicks "Save"
4. a. If the form is not valid, we show errors (and the flow starts from 2.)
    b. If the form is valid, we ask the user for authentication
5. We subscribe to `newCommunity`
6. We wait until the subscription port sends us something back
7. a. If it sends "responded", redirect to the community (and the flow is over)
    b. If it sends "starting" we send the Eos transaction
8. a. If the Eos response was valid, redirect to community (we should remove this, so we only redirect when the subscription tells us to)
    b. Otherwise, show errors and start from 2.
    
So the old flow validates the model very late (after subscribing to `newCommunity` and asking the user for authentication - IMO it should be before those), while the new one validates it as soon as the user clicks "Save". Because of that, we have the validated information on the `SubmittedForm` msg, not only just before sending the Eos transaction. So we didn't need the decoder before because we only validated the data when sending the Eos transaction - which means it would take longer to show errors to the user.

The result ends up being the same, but the process is a little different, and better for the user. By the way, `minBalance` already was in `CreateCommunityData`, it just wasn't included in the encoder. I'm still convinced `minBalance` should be in `CreateCommunityData` - that's an abstraction we make throughout the app (we don't have a `Community` and a `Token`, the token information is already included in `Community`). Having `min_balance` in the encoder doesn't seem to break anything, but we could make it so the encoder actually returns two encoders (in a tuple), one for the community and one for the token - I think this makes sense, because when we're creating a new community, we're also creating a new token anyway

NeoVier:
@lucca65 I've added a flag to disable using subdomains. It defaults to true, but you can do `USE_SUBDOMAIN=false yarn start` to disable it and make the app behave as before

src/elm/Page/Community/Settings/Currency.elm
lucca65: those are community settings, not token settings... Maybe all of this confusion and discussion is because Figma is wrong? 

because if the criteria for adding invitations to the token is because it envolves some type of reward, then objectives should also be there.... 

The community screen should include the `inviterReward` and `invitationReward` those are related to stuff that happens on the `network` of members, tokens can still exist outside of it and need to be like that in order for it to be compatible with other tools (like wallets).

src/elm/Community.elm
lucca65: bro, first of all kudos to the complete and amazing explanation. I see this as a good example to be followed and thanks for being this reference!

I've loved the new flow. It is indeed much better. I wonder how we could use Elm to define state machines like this. Anyway...

You've mentioned something that I wasn't aware of: that its possible to have the `responded` reply if the community is already created. This means that we could use other stuff like `transaction_id` in place of `symbol`. This would give us a generic subscription that guarantees transaction pushing and processing to the frontend.  Because now we have multiple subscriptions (creatingCommunity and transfers are the ones I remember). With your confirmation it means it's possible to us to create this. [I've opened an issue on the backend](https://github.com/cambiatus/backend/issues/148), might change the whole responsiveness of the app!

> Having min_balance in the encoder doesn't seem to break anything, but we could make it so the encoder actually returns two encoders (in a tuple), one for the community and one for the token - I think this makes sense, because when we're creating a new community, we're also creating a new token anyway

About this, I think this the right direction, adding the token data to that type. I still prefer to have one type for creating community and another for token, and treat them differently. Messing them both up will lead to an even confusing codebase because the frontend will have a different view of what an community is than on the blockchain. Changes in the contract may hurt a lot in the future if the frontend is not on par with it

NeoVier: > I wonder how we could use Elm to define state machines like this.

To me, Elm is pretty much exactly like a state machine (which is a topic I really really like - In fact my first ever Elm program was a [state machine simulator](https://neovier.github.io/LFC01/) - plus some other stuff - for a class project. You can see the source [here](https://github.com/NeoVier/LFC01) if you want. In there, there are some example input files as well), in which you have a `Model` that represents states and `Msg`s that represent transitions. Anyway...

> You've mentioned something that I wasn't aware of: that its possible to have the responded reply if the community is already created. This means that we could use other stuff like transaction_id in place of symbol. This would give us a generic subscription that guarantees transaction pushing and processing to the frontend. Because now we have multiple subscriptions (creatingCommunity and transfers are the ones I remember). With your confirmation it means it's possible to us to create this. I've opened an issue on the backend, might change the whole responsiveness of the app!

That's an amazing accidental finding! :sweat_smile: Tbh, I'm not really sure if that response is even used by the backend right now, but I see how this could be nice, and help with things like [this](https://github.com/cambiatus/frontend/blob/feature/support-subdomains/src/elm/Page/Community/ActionEditor.elm#L1060-L1061) and [this](https://github.com/cambiatus/frontend/blob/feature/support-subdomains/src/elm/Page/Community/ObjectiveEditor.elm#L382-L383).

> About this, I think this the right direction, adding the token data to that type. I still prefer to have one type for creating community and another for token, and treat them differently. Messing them both up will lead to an even confusing codebase because the frontend will have a different view of what an community is than on the blockchain. Changes in the contract may hurt a lot in the future if the frontend is not on par with it

Yes, I guess separating these appropriately can prevent future headaches. As this is a "lower layer" inside the frontend (it's closer to the blockchain than to the user), I think it makes sense to follow the way we're doing it in the blockchain. The `Community.Model` is closer to the user, so there it makes sense to have the community and the token data together - it's an abstraction the user does. In their point of view, the community and the token are kind of the same thing. Let's do it! :rocket:

src/elm/Page/Community/Settings/Currency.elm
NeoVier: > those are community settings, not token settings

From our point of view, yes, I agree. However, from the user's point of view, they probably don't even know there is a token separate from the community - it's all the same thing. For them, a community has a name, a description, a currency, a logo, etc.

So the user's separation of concerns is very different from ours. We know there are two entities - a community and a token - but they don't necessarily know that (AFAIK). The way it is now, we have two screens:
- "Community Information": this is used to manage what the community "looks" like - name, logo, description, etc.
- "Currency and rewards": this is used to manage how the community works - invited reward, inviter reward and minimum balance (following this, maybe the auto invites should be in this screen as well? But IMO it doesn't really fell right in this screen)

Also, moving these two fields to "Community Information" would lead to a pretty big form there, and the "Currency and rewards" screen would have a form with a single input field. Maybe we can talk to the design team about this? @heltonlr

src/elm/Page/Community/Settings/Currency.elm
lucca65: @heltonlr ping

src/elm/Page/Community/Settings/Currency.elm
lucca65: > So the user's separation of concerns is very different from ours. 

That's exactly the problem, I don't see why it should be that way.

> Also, moving these two fields to "Community Information" would lead to a pretty big form there, and the "Currency and rewards" screen would have a form with a single input field. 

Not true my bro, I urge you to take a look at the [token contract](https://local.bloks.io/account/cambiatus.tk?nodeUrl=http%3A%2F%2Fstaging.cambiatus.io&coreSymbol=SYS&systemDomain=eosio&loadContract=true&tab=Tables&account=cambiatus.tk&scope=cambiatus.tk&limit=100&table=expiryopts)!

A list of information related to Token:

- Current Supply (informative only)
- Max possible supply (editable)
- Min balance (editable)
- Issuer (editable -- but I think we shouldn't allow it for now)
- type (can be expiry or mcc)

Tokens can have multiple types, you are already familiar with configurations for the `mcc` one -- mutual credit clearing. But the token can be of type `expiry` that will come with more fields:

- Natural person expiration period in seconds (editable)
- Juridical person expiration period in seconds(editable)
- Renovation amount (editable)
- currency (the SYMBOL used -- view only)

So that's 10 fields. I won't count but it surely have a lot. Well this conversion is interesting because it shows how little we use from the structure that we already have. The `/community/new` is super simple and very old (you must have noticed from the code) and that shows how much we can improve there! Maybe the designs are not ideal not even considering that it has so much stuff.

I keep my position of having two screens, and not creating an different abstraction for the user, keep the one in the contract! that initial screen that kept both together was from a time that we didn't even had the `expiry` token type and a lot has changed!

src/elm/Page/Community/Settings/Currency.elm
heltonlr: Ping @rafachadud

heltonlr: Oi @lucca65 , the number 3 is an example of what a value looks like in the configuration that the user made, in the case of 3 MIZO (which has 0 decimal places). What were you confused about?

src/elm/Page/Community/Settings/Currency.elm
NeoVier: > Not true my bro, I urge you to take a look at the token contract!

Oh, I had no idea we had so many possible fields for the token!

In that case, I guess we should rename from "Currency and rewards" to something more related to token?

> you are already familiar with configurations for the mcc one -- mutual credit clearing.

Just to be sure: `mcc` is how most (all?) communities work right now, and doesn't introduce any other fields, right?

> The /community/new is super simple and very old (you must have noticed from the code) and that shows how much we can improve there!

Indeed. I think we should tackle that in another issue though - I created #514 for this

src/elm/Page/Community/Settings/Currency.elm
lucca65: Yeah I know! I mean we are already super familiar with the whole flow and we know about precision! What I'm trying to say is that this screen configured with 0 don't gives the idea that this is a token without decimal cases. Maybe we can show pi but only highlight the available decimal cases (in this case 0)

src/elm/Page/Community/Settings/Currency.elm
lucca65: > In that case, I guess we should rename from "Currency and rewards" to something more related to token?

Yeah I do think this screen needs improvements, it doesn't make sense that we update the whole screen to work with 
Cambiatus' features from one year ago!

> Just to be sure: mcc is how most (all?) communities work right now, and doesn't introduce any other fields, right?
> 
> I wonder if we can make these changes right away without breaking anything on other parts of the frontend or the backend (I suppose we can, but just want to be sure). Also, can a community freely change between mcc and expiry whenever they want or is there some sort of restriction around that?

There is only one community that uses `expiry`, Ticolones (which is not active). A community will not be able to change between token strategies once it was created. Maybe we can support something like this in the future but the contract doesn't allow that now

NeoVier: @rafachadud can you take a look at this as well? We're moving the "inviter reward" and "invited reward" fields to the "Community information" page, and adding some fields to the "Currency and rewards" page as [listed above](https://github.com/cambiatus/frontend/pull/494#discussion_r617734936). Also, the "Currency and rewards" page should probably be renamed to something like "Token information" (you can read the reasoning behind these changes in this thread)

src/elm/Community.elm
lucca65: > To me, Elm is pretty much exactly like a state machine (which is a topic I really really like - In fact my first ever Elm program was a state machine simulator - plus some other stuff - for a class project. You can see the source here if you want. In there, there are some example input files as well), in which you have a Model that represents states and Msgs that represent transitions. Anyway...

Nice bro! I think we could improve our Msgs to be more explicit, right now we need to check the `update` function to know what happens next... Just throwing out ideas to have more expressive flows on our pages

>That's an amazing accidental finding! 😅 Tbh, I'm not really sure if that response is even used by the backend right now, but I see how this could be nice, and help with things like this and this.

I've moved [this issue](https://github.com/cambiatus/backend/issues/148) from improvements proposals, we can work on it when possible!

src/elm/Page/Community/New.elm
```diff
@@ -1,4 +1,4 @@
-module Page.Community.Editor exposing
+module Page.Community.New exposing
```
lucca65, line 1: nice!

lucca65 (changes requested):
I'm getting this blank screen on chrome: 
Captura de Tela 2021-04-26 às 10 30 57

NeoVier:
That blank screen is because of the changes in the GraphQL in the backend (you can see the error in the console if you have it open before the page loads, but it's failing because we don't have `precision` as a field anymore), I'll update the GraphQL types

src/elm/Api.elm
```diff
@@ -5,6 +5,7 @@ module Api exposing
     , communityInvite
     , get
     , getBalances
+    , getExpiryOpts
     , getTableRows
     , signInInvitation
```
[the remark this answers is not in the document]
NeoVier: Yeah, I was actually looking at [this post](https://jfmengels.net/simplify/#automatic-fixes-for-unused-dependencies) by the creator of elm-review, and was thinking we could maybe use some more elm-review rules (it should be on a different issue though). You can see for yourself by running

```sh
npx elm-review --template jfmengels/elm-review-unused/example
```

Well... I get 716 errors in 107 files, just of unused stuff. As you have already mentioned, we have a lot of old code, and I think this would really help with that, what do you think?

src/elm/Community.elm
```diff
@@ -116,6 +118,8 @@ type alias Model =
     , hasKyc : Bool
     , hasAutoInvite : Bool
     , validators : List Eos.Name
+    , coverPhoto : Maybe String
```
lucca65, line 121: This should be a list

src/elm/Page/Community/New.elm
```diff
@@ -502,13 +512,14 @@ validateModel accountName model =
                         , logoUrl = logoUrl
                         , name = name
                         , description = model.description
-                        , subdomain = Just model.subdomain
+                        , subdomain = model.subdomain ++ ".cambiatus.io"
```
lucca65, line 515: one day we will allow communities to go full whitelabel and use their own domains!

src/elm/Api.elm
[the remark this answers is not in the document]
lucca65: for sure bro, the frontend does need a lot of clean up! this can help a lot!

src/elm/Page/Community/New.elm
NeoVier: Should we account for that right now though?

src/elm/Community.elm
NeoVier: Yep, I fixed it today, just haven't pushed it yet

src/elm/Page/Community/New.elm
lucca65: no, not at all! just a comment (:

NeoVier:
@heltonlr since this is a really big one, you could start testing it out already! The only thing missing is using the auto invitations (so right now you won't be able to join a community through the "Join" page), but everything else should be working! (I was just giving it another test and found a bug when creating a community, so that's also not working right now)

For some reason there isn't a netlify deploy for this branch :thinking: so I deployed it to staging.

Note: if you try to access just staging.cambiatus.io and you were logged out before, you'll be redirected to cambiatus.com (because you're a guest trying to access a non-existing community), so you'll either need an invitation to a community or try to get into one that has auto invites (you can use mine: https://buss.staging.cambiatus.io/, but you'll still need an account that already exists). Thinking about it, this sounds kind of weird, is this the expected behavior or should we do something else?

heltonlr:
Hey Team, some points that I find during the tests:

- All-time that I use the community selector I am dislodged and redirected to the 12 words screen'
- When I am navigating in a community and I use the notifications and click on a notification from another community I am redirected to that community but the Logo on the header remains of the previous community. The correct is to change the Logo like Slack;

Is that for now, I am testing yet.

heltonlr:
I created the community FINANCE, I try to access using the community selector and that error occurs

https://user-images.githubusercontent.com/39558132/117898384-3ee80480-b29b-11eb-84f6-f8bf0694c00f.mov

...

heltonlr:
Nessa tela nós podemos deixar o logo oficial da Cambiatus, já que é o Logo do produto, da nossa empresa...sim, é o mesmo da comunidade Cambiatus mas ok, futuramente podemos alterar o Logo da comunidade Cambiatus para não confundir...

Screenshot_20210511-211318_Chrome

NeoVier:
@heltonlr thanks for testing! I'll address these issues more carefully tomorrow, but here's some initial information:

> When I am navigating in a community and I use the notifications and click on a notification from another community I am redirected to that community but the Logo on the header remains of the previous community. The correct is to change the Logo like Slack;

Actually, you're still on the same community (you can see the URL isn't changing when you click a notification). I'll make it so you're redirected to the appropriate community!

> I created the community FINANCE, I try to access using the community selector and that error occurs

Right now there's an issue that when creating a community, it's subdomain isn't being saved on the blockchain, so that's why you're being redirected to `.staging.cambiatus.io` (as if the subdomain was nothing). @lucca65 this was [the issue I mentioned on Slack](https://cambiatus.slack.com/archives/CA83HJAAD/p1620419479063400) about community creation

env-config.js.sample
```diff
@@ -4,6 +4,7 @@ window._env_ = {
   CHAIN_ID: 'fa087d6c692f16e01a9864749829359cd26b48db703377893f32ff1c72673a78',
   API_URL: 'https://staging.cambiatus.io',
   GRAPHQL_URL: 'https://staging.cambiatus.io/api/graph',
+  GLOBALSTORAGE_URL: 'https://staging.cambiatus.io/globalstorage',
```
lucca65, line 7: we should put USE_SUBDOMAIN as well

public/translations/en-US.json
```diff
@@ -883,11 +895,72 @@
...
-      "description": "Logo, name, description"
+      "description": "Logo, name, description, url",
+      "guidance": "Guidance",
+      "logo": {
+        "title": "Community logo",
+        "description": "Your file must be in PNG, JPG or SVG format. It should measure 80px in width and 80px in height to be better viewed on the interface."
+      },
+      "invitation": {
+        "title": "Access to the community",
+        "description": "When a community \"invitation\" is set, new members only get access through an invitation."
```
lucca65, line 906: I wonder if this could be more well phrased. Maybe something like:

"When a community is invite only" or something like this. The way it is looks like a long round trip to get to this meaning

public/translations/en-US.json
```diff
@@ -883,11 +895,72 @@
...
+      },
+      "cover_photo": {
+        "title": "Community cover picture",
+        "description": "Be sure to add a picture that has good quality"
+      },
+      "fields": {
+        "name": "Community name",
+        "description": "What is the community",
```
lucca65, line 914: maybe "what this community is about" or "Describe this community"

lucca65: what gives a weird meaning, in portuguese it would be something like "o que é a comunidade?", idk know sounds bad

public/translations/en-US.json
```diff
@@ -883,11 +895,72 @@
...
+        "website": "Community's website",
+        "url": "Community custom URL",
+        "invitation": "Invitation"
+      },
+      "placeholders": {
+        "name": "Give your community a name",
+        "description": "Say something about your community",
```
lucca65, line 921: not sure about the "your community", looses the sense of collective

src/elm/Api.elm
```diff
@@ -36,13 +30,6 @@ backendUrl { endpoints } paths queryParams =
         queryParams
 
 
-backendUrl_ : Endpoints -> List String -> List QueryParameter -> String
-backendUrl_ endpoints paths queryParams =
```
lucca65, line 40: I'm not sure about this but to upload images and create invites don't we need this?

src/elm/Api/Graphql.elm
```diff
@@ -19,17 +19,24 @@ withAuthToken authToken =
...
+
+
 query : Shared -> Maybe String -> SelectionSet a RootQuery -> (RemoteData (Graphql.Http.Error a) a -> msg) -> Cmd msg
-query { endpoints } maybeAuthToken query_ toMsg =
+query ({ endpoints } as shared) maybeAuthToken query_ toMsg =
     query_
         |> Graphql.Http.queryRequest endpoints.graphql
+        |> withCommunityDomain shared
         |> withAuthToken maybeAuthToken
         |> Graphql.Http.send (RemoteData.fromResult >> toMsg)
```
lucca65, line 27: this code is looking 🍠

src/elm/Community.elm
```diff
@@ -142,31 +141,20 @@ communitiesSelectionSet =
...
-
-
 communitySelectionSet : SelectionSet Model Cambiatus.Object.Community
 communitySelectionSet =
     SelectionSet.succeed Model
         |> with Community.name
         |> with Community.description
         |> with (Eos.symbolSelectionSet Community.symbol)
         |> with Community.logo
+        |> with (Community.subdomain Subdomain.name |> SelectionSet.map (Maybe.withDefault ""))
```
lucca65, line 151: oh, this is a maybe? not good, subdomain.name can't be null. I'll fix the type

src/elm/Eos.elm
```diff
@@ -312,7 +342,7 @@ symbolFromString str =
     in
     case ( maybeSymbolCode, maybePrecision ) of
         ( Just symbolCode, Just precision ) ->
-            if String.length symbolCode == 3 || String.length symbolCode == 4 then
+            if String.all Char.isAlpha symbolCode && (String.length symbolCode == 3 || String.length symbolCode == 4) then
```
lucca65, line 345: oh right, good, since we are here, the actual max length for symbols is 7

lucca65 (commented):
Are we using the loading.gif? the svg is enough, isn't it?

lucca65:
1. When clicking on a notification of a different community I'm in, it asks for my 12 words after the redirect. I've opened a fresh session, logged in to MIZU and tried to go to a transfer for CMB, after the second login, i could go back and forth without having to provide the 12 words again. This happens at every different community I try
2. The transfer links on the dashboard don't redirect me to the community
3. I always get the error message: "we can't find expiration options for your token". this should only be an error if the token type is expiry (contract `cambiatus.tk`, table `stat`, scope is the symbol_code `MIZU`, `CMB`. No precision needed

src/elm/Api.elm
NeoVier: There's a `backendUrl` function right above this one that is exactly the same, except it takes an entire `Shared` instead of just the `Endpoints`

src/elm/Community.elm
NeoVier: I thought you said we needed to deploy the update first :thinking:

src/elm/Api.elm
lucca65: oh okay

src/elm/Eos.elm
NeoVier: Can we safely change that here or is there somewhere else we need to change as well?

src/elm/Community.elm
lucca65: we do, but they are different!

the one i've mentioned before is to make all communities have an obligatory subdomínio, but here is about the name not being null

lucca65: or am i reading this code wrong? 🤔 maybe so..

this is like this because the name can be null right?

NeoVier: Oh, right, makes sense

NeoVier:
> Are we using the loading.gif?

We aren't, and it's deleted already

I'll take a closer look to the other points later today!

src/elm/Eos.elm
lucca65: not sure how we parse symbols on the frontend, but probably we will have no problems

src/elm/Community.elm
NeoVier: Actually, subdomain can be null, but subdomain.name is always a string:

2021-05-18-083531_461x85_scrot

lucca65: oh ok, i've misread the code then

heltonlr:
So, I think I finished the tests:

- I tried to join a community that is open for new members
**Success**

- I tried to join a community that is close for new members (only invite)
**Success**

- I tried to invite a new member to both (no-invite and only-invite communities)
**Success**

- I tried to change a community that was invite-only to no-invite
**Success**

- I tried to change the community by selector on the header
**Success, but sometimes I was redirected to 12 words screen**

- I tried to access a notification of another community (no selected on a select element on the header)
**Success, all the dashboard changed**

- I tried the communities' logo
**Success**

- I tried the communities' image background
**Fail**

- I tried the communities' website
**Fail**

- I tried the communities' name
**Success**

- I tried the communities' URL
**Success**

@lucca65 @NeoVier 

Would you like that me test some other scenarios?

NeoVier:
> I tried the communities' website
Fail (The button didn't open the website)

@heltonlr can you elaborate on this? I can't seem to replicate it. Are you setting the website on the "New community" page or on the "Community information" page? The button shows up, but when you click it, it doesn't take you to the community's website?

heltonlr:
> > I tried the communities' website
> > Fail (The button didn't open the website)
> 
> @heltonlr can you elaborate on this? I can't seem to replicate it. Are you setting the website on the "New community" page or on the "Community information" page? The button shows up, but when you click it, it doesn't take you to the community's website?

I edited a community, I set the external website (ex.: HTTP://muda.com.br) and when I tried to click on the visit website's button, an error occurs ("Page not founded").

lucca65:
@heltonlr please remember to don't use HTTP uppercase. Also Muda's website is http://muda-oe.com

NeoVier:
@heltonlr if you manually go to the website you entered, does it work? I think you're probably using an invalid website

NeoVier:
> I always get the error message: "we can't find expiration options for your token". this should only be an error if the token type is expiry (contract cambiatus.tk, table stat, scope is the symbol_code MIZU, CMB. No precision needed

@lucca65 when do you get this error? Is it on the console or as a feedback?

lucca65:
as a feedback @NeoVier

NeoVier:
Ok @lucca65 apparently your community (I'm assuming it's happening with `0,MIZU`) is set as `expiry` instead of `mcc`, but it doesn't have an entry in the `expiryopts` table, so it obviously errors out when trying to get data from that table. Maybe this was some configuration problem when you created the community? :thinking:

NeoVier:
> When clicking on a notification of a different community I'm in, it asks for my 12 words after the redirect. I've opened a fresh session, logged in to MIZU and tried to go to a transfer for CMB, after the second login, i could go back and forth without having to provide the 12 words again. This happens at every different community I try

Damn, I really can't replicate this :grimacing:. The way the app starts up right now is:

1. Check if the url ends with `/globalstorage`. If so, execute the app as globalstorage. If not, keep going.
2. Check if that specific domain has synced with globalstorage at least once (this is done via a localStorage flag that's stored only on the subdomain, not on globalStorage).
  2.1. If the domain hasn't been synced and there is user information on the domain's localStorage, send all relevant keys to globalStorage and overwrite anything there
  2.2. If the domain has been synced, or there isn't any user information on the domain's localStorage, get all relevant keys from globalStorage
3. Set the flag that determines if the domain has been synced with globalStorage as `true`
4. Run the app as usual, but using globalStorage instead of localStorage

(By user information I mean the `USER_KEY`)
I can't really think of other edge cases that you might be falling into... These steps all happen before we even start the Elm app.

Can you check on the console if there is some logging that says you've been logged out? (It would probably be shown as `logout port started` or something like that). You can also check the `storage` tab on the developer console and check what's being stored in the domain's localStorage and what's being stored in the `staging.cambiatus.io` domain

heltonlr:
> @heltonlr if you manually go to the website you entered, does it work? I think you're probably using an invalid website

Ok, I tried using a valid url now. Works well

lucca65:
> Ok @lucca65 apparently your community (I'm assuming it's happening with 0,MIZU) is set as expiry instead of mcc, but it doesn't have an entry in the expiryopts table, so it obviously errors out when trying to get data from that table. Maybe this was some configuration problem when you created the community? 🤔

oh. sorry then bro, my bad, I should have checked it before hand. it won't happen again 💪

lucca65:
I must suppose then its something on Safari. I've tested it on Safari, Chrome and Firefox and it only happens on Safari consistently. Which is a problem because we must support Apple ecosystem, as it is almost 20% of our user base. We have a few alternatives: 

1. VM your way into a windows and install safari
2. get an iPod touch / iPhone and test it calling it to your machine
3. try [Browserstack](https://www.browserstack.com)

Either way we need to support it bro ):

lucca65 (approved):
🍬

NeoVier:
@lucca65 Bad news on the safari support... After a lot of testing and digging, I stumbled upon [this StackOverflow thread](https://stackoverflow.com/questions/63922558/safari-localstorage-not-shared-between-iframes-hosted-on-same-domain), which is pretty much exactly what we're trying to do. Safari introduced a security feature where we can't do this (with localStorage), as far as I understand. I've tried using `globalstorage.staging.cambiatus.io/globalstorage` (so it's considered first-party), and it also didn't work.

The option introduced in [the Storage Access API](https://webkit.org/blog/8124/introducing-storage-access-api/) is to use cookies instead of localStorage, which I really don't like, but it seems to be our only option to support the latest safari releases :grimacing:. What do you think we should do?

lucca65:
damn... really? 😢

lucca65:
This is so bad, I needed a time to digest this. "Aos 45 do segundo tempo" 😩

Its bad that advertises used this feature to fuck privacy up. Damn Well, we have no choice and we should be more attentive about Apple's ecosystem. Safari is the new Internet Explorer

NeoVier:
Should we try using cookies then?

lucca65:
we have no other alternative, right?

lucca65:
yeah, I'll let the team know this is going to take longer yet ):

NeoVier:
@lucca65 everything should work now! We're using cookies, and moving all the localStorage data people might have to cookies as well, so they're not logged out as soon as the update is released.

Also, if we want to keep sessions between communities when developing on localhost, we should use urls like `cambiatus.staging.localhost:3000`, because [domain names must have at least 2 dots](https://stackoverflow.com/questions/1134290/cookies-on-localhost-with-explicit-domain#1188145), but we should automatically be redirected to that url when trying to go to `localhost:3000` anyway, so it's no big deal!

lucca65 (approved):

PR #499, opened 2021-03-30
Fix/v1.9.0 hotfix

### LINKED ISSUE #498: Show out of stock on search
Users are frustrated that when searching for an item, they only know its out of stock after they clicked it.

Show out of stock on search

Show the full card as seem on the shop page, as users are already used to them

**Additional context**
Suggested by Jannelle from Verdes

lucca65:
closes #498

Some small fixes that appeared after our [last release](https://github.com/cambiatus/frontend/releases/tag/v1.9.0):

- Remove animations from Feedback
- Show out of stock on search results 
- Allow sorting direction to be chosen 

1. Try to hover in any feedback, it shouldn't zoom in anymore
2. Search for an item that is out of stock. It should show in the listing as Out of Stock, without showing the price
3. Go to the analysis page, click the sorting button and make sure the ordering of the claims is using date to sort it correctly


src/elm/Page/Dashboard/Analysis.elm
```diff
@@ -595,15 +636,13 @@ fetchAnalysis { selectedCommunity, shared, authToken } { profile, statusFilter }
...
-                ( Absent, Absent ) ->
-                    Absent
+            , direction =
+                if direction == ASC then
+                    Present Cambiatus.Enum.Direction.Asc
 
-                ( _, _ ) ->
-                    Present filterRecord
+                else
+                    Present Cambiatus.Enum.Direction.Desc
```
NeoVier, line 644: Using a `case` statement here would be more appropriate (personally, it always seems kind of wrong to compare type constructors like this). For now, we're only sorting ascending and descending, but if someday we decide to sort by pending/accepted for example, it would be easier to know where to change it, and we would avoid possible weird bugs

src/elm/Page/Dashboard/Analysis.elm
```diff
@@ -566,9 +584,32 @@ update msg model loggedIn =
...
+                newModel =
+                    { model
+                        | filters = { oldFilters | direction = sortDirection }
+                        , reloadOnNextQuery = True
+                        , status = Loading
+                    }
+            in
+            newModel
+                |> UR.init
+                |> UR.addCmd (fetchAnalysis loggedIn newModel.filters Nothing)
```
NeoVier, line 608: Is there a reason we can't just `List.reverse` the claims? That would mean less loading times and better UX, and I don't think we get other benefits from fetching everything again, other than maybe fetching new claims.

Of course, if we ever need to sort by some other parameters, we could create a function like
```elm
applySort : FilterDirection -> List Claim.Model -> List Claim.Model
```
but I guess that's premature optimization right now :sweat_smile:

NeoVier (approved):
Looks pretty good, just have a couple minor pointers, let me know if you agree with them and if you think it's worth addressing them

src/elm/Page/Dashboard/Analysis.elm
lucca65: nice! I'll change it!

src/elm/Page/Dashboard/Analysis.elm
lucca65: yeah, the problem is that this is a query that is paginated. Imagine we have 100 claims, the older one from December 2019 and the newest March 2021. If we did the reversing on the frontend like you described we would only reverse the first batch we queried (today is 16). To see easily the claim from 2019 we would need to click the "more" button a couple of times until we had all the claims loaded and then reverse it.

This implementation is done by the backend to allow validators to catch on with old claims that simply were left behind. Some users do like 10, 15 claims each days, so its easy to some to get behind. 

that's why its an API call to reverse, instead of a local one

In the shop the picture is different because there we dump the whole thing

NeoVier: Makes sense, I didn't take into account that this was paginated

NeoVier (approved):
Awesome :muscle:

PR #502, opened 2021-03-31
Fix/auth page improvements

### LINKED ISSUE #501: Login page/auth improvements
The login page has a couple of big problems:
- Some people, on smaller devices, can't see the error messages
- If there is an error after submitting, the user needs to reload the page to try again

Also, on the developer point of view, the `Auth` module is very difficult to grasp, and has a bunch of (hidden) unused things (like some of the `Status` constructors) which makes it a lot harder to understand

Better error handling on the login page, and some simplification on the developer point of view

(Re)Use a view like the `LoggedIn`'s `Feedback` system to display error messages

**Additional context**
Originated from [this slack thread](https://cambiatus.slack.com/archives/CA83HJAAD/p1617114363142400)

NeoVier:
Closes #501 

A lot of the changes in this PR are transparent to the user.

What the users can see:
- `Guest` pages (Login and Register) now use the same `Feedback` system as `LoggedIn` pages (the red/green bar at the top of the screen).
- The `Feedback` styles have been fixed so the bar sticks to the top once the user scrolls past it, so it's always visible.
- The Login page has a better flow. If an error happens, the user doesn't need to reload the page to try again.
- For browsers that support the [clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) (for example, chrome for desktop), there is now a "Paste" button so users can paste in their 12 words in the Login page.

What the developers can see:
- The `Auth` module is much smaller and easier to understand. It's now only concerned about logged in users, and can be in a `WithoutPrivateKey` or `WithPrivateKey` state.
- The logic and view of the `Auth` module that dealt with non-logged in users is now in the `Login` module, since that was the only page that used it.
- The `View.Input` module allows for more flexibility and customization on counting strategies, counter styles, error styles, HTML elements inside the input area, and using different HTML elements as input (such as `input` or `textarea`).
- The `Pin` module is now a component with its own model/view/update, so it can be used both by `Guest` and `LoggedIn` without much code repetition.
- Much more consistent logging when dealing with ports and JS.
- Some extra logs through the authentication/login process.

This deals with authentication, so it's important to be thorough!

The main affected area is the login page, so try logging in with valid and invalid data, check if the errors are easily seen on any screen size (mainly that they stick to the top of the screen if you've scrolled down enough). Make sure you can still register a new user as well.

This also deals with the PIN modal, so try doing something that requires that, and see if it behaves accordingly.

Also check if the feedback bar at the top of the screen works as expected (is always visible even after scrolling past it) on the pages where you need to be logged in to see.


src/elm/Auth.elm
```diff
@@ -230,23 +233,18 @@ type alias PrivateKey =
...
 isAuth model =
     case model.status of
-        LoggedInWithPin _ ->
-            True
-
-        LoggedInWithPrivateKey _ ->
+        WithPrivateKey _ ->
             True
 
+        -- TODO - Should `WithoutPrivateKey` be considered authed here? (Check usage in other modules)
```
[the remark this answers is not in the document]
NeoVier: Yeah, semantically it makes sense to consider the user authenticated, just need to make sure the modules that use this function use it accordingly (before this change, only the states that had a `PrivateKey` were considered authenticated). 

Maybe renaming it would be more appropriate, since other modules basically use this to check if they need to call `LoggedIn.RequiredAuthentication`, which asks the user for their PIN.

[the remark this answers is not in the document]
NeoVier: Also, thinking about it, I don't really like the structure we have right now. We're kind of duplicating the authentication state with `Guest`/`LoggedIn` and the `Auth` module. As far as I can tell, a `Guest` can only ever have an `Auth.Model` that is `Unauthenticated`, and (if I'm not mistaken) that's only used in the `Login` page (which at least in this PR is just a blank page that wraps around `Auth`). Similarly, a `LoggedIn` user is always either `WithPrivateKey` or `WithoutPrivateKey`, and the view for those is completely different from `Unauthenticated`. The only thing that's shared between the authenticated and unauthenticated states is input validation, input messages and a few helper functions.

Not sure what the best solution is, but one approach would be to get rid of the `Auth` module (at least the data modelling part, we could still reuse the validation and the views), and change the `Auth.Model` for a `Maybe PrivateKey` field in `LoggedIn.Model`. IMO that would be an equivalent model to the one we have now, but a lot easier to understand, especially to new people that aren't familiar with the way we're doing authentication (it actually took me a while to realize how we were getting the `PrivateKey` :sweat_smile:).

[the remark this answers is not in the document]
lucca65: hehehe I love it bro!

Ok, lets do it!

## Old Context
At some point the app was very flexible and allowed for a very different usage than we have right now. It allow any user to use the app freely without having a private key, only asking for it when it needed signing a transaction (we were backend free by then). It also supported some sort of login in, it was not as we have it today, as it only needed to have _something_ in the local storage to be considered logged, no private key required. With that "middle" state of logged without a private key you could do anything, except sign a transaction. 

You could also have a variance of that, allowing the key to be stored by setting a PIN, or adding it only to the next transaction (it would be discarded afterwards). It even supported Guest pages, public pages that could be seem by anyone. Community screen was one of those screens.

## New context
Ok, history class over. The app change drastically to have a backend, and now we **require** backend authentication and require PIN and private key to be able to use the app. 

Effectively we have some pages that can be seem as a guest (maybe we can call them public and restricted instead of guest and logged)

- Payments page
- Invitation pages
- Register pages
- Sign in page

all the rest are restricted. 

## New requirements
I'm ok on us removing Auth, but we need to actually have more info than just the private key. Private key is the base but we also need a way to know what's up with the authentication token. Check the validity of the token and only progress if it is ok. In every full page load, and even periodically we will need to check this. Imagine the app hasn't checked the auth token in a while, or got a 403 from the backend saying the token isn't valid anymore while doing some other requests. The app will have to know this, and using the private key renew the auth token. All of this need to be transparent to the user.

I agree with you @NeoVier that this part of the app is obscure, over complicated and a big technical debt that we need to sort out. The tradeoffs are incredible and necessary at this point!

Changing to this strategy also makes easier for us to use Natus' approach to authentication, based off email, creating users and relating them to one private key and one account, that can be changed and recovered.

src/elm/View/Form/Input.elm
```diff
@@ -82,17 +85,25 @@ toHtml options =
...
+             ]
+                ++ options.extraAttrs
+            )
+            []
+        , case options.extraElement of
+            Nothing ->
+                text ""
+
+            Just element ->
+                element
```
lucca65, line 105: this could be `Maybe.withDefault`

lucca65 (commented):
Just some commenting on it! Keep up bro, you are doing good! 🏃‍♂️ 🔥

public/translations/cat-CAT.json
```diff
@@ -941,4 +948,4 @@
       }
     }
   }
-}
+}
```
lucca65, line 951: Add LF file endings on your editor to avoid this

public/translations/en-US.json
```diff
@@ -104,6 +104,10 @@
     "invalidSymbol": "Invalid symbol.",
     "unknown": "Uoh. Something wrong happened.",
     "accountNotFound": "Account not found. Are you sure you created your account?",
+    "accountDoesNotCorrespond": "This private key does not correspond to logged in account. Please logout to use a different account.",
```
lucca65, line 107: If we ever got to this state, we cannot do nothing about it, the session is doomed, maybe we can show a modal displaying this message and force a logout?

src/elm/Session/LoggedIn.elm
```diff
@@ -205,25 +197,15 @@ initModel shared authModel accountName selectedCommunity authToken =
...
     = Loading Eos.Name
     | LoadingFailed Eos.Name (Graphql.Http.Error (Maybe Profile.Model))
     | Loaded Profile.Model
 
 
-isAuth : Model -> Bool
-isAuth model =
-    Auth.isAuth model.auth
+hasPrivateKey : Model -> Bool
+hasPrivateKey model =
```
lucca65, line 207: so about this rename, does it make sense we consider if we have the private key? if we are logged we need to have it 100% sure. We will need the auth token and private key

src/elm/Session/LoggedIn.elm
```diff
@@ -177,8 +169,8 @@ type FeatureStatus
     | FeatureLoading
 
 
-initModel : Shared -> Auth.Model -> Eos.Name -> Symbol -> String -> Model
-initModel shared authModel accountName selectedCommunity authToken =
+initModel : Shared -> Maybe String -> Eos.Name -> Symbol -> String -> Model
+initModel shared maybePrivateKey_ accountName selectedCommunity authToken =
```
lucca65, line 173: we could introduce a type for private key instead of using a `String`

src/elm/Session/LoggedIn.elm
```diff
@@ -489,7 +436,7 @@ viewHeader ({ shared } as model) profile_ =
                   else
                     text ""
                 ]
-            , div [ class "relative z-20" ]
+            , div [ class "relative z-50" ]
```
lucca65, line 439: lol

src/elm/View/Feedback.elm
```diff
@@ -0,0 +1,69 @@
...
+
+
+type Status
+    = Success
+    | Failure
+
+
+type Model
+    = Hidden
+    | Shown Status String
```
lucca65, line 20: Shown gives the idea that its something on the past, not on the present, maybe we can use `Visible` to be opposed to `Hidden`?

src/index.js
```diff
@@ -83,6 +82,10 @@ function getUserLanguage () {
   )
 }
 
+function canReadClipboard () {
+  return !!navigator.clipboard && !!navigator.clipboard.readText
```
lucca65, line 86: 😞 javascript

lucca65 (changes requested):
On safari for Mac this happens when I click the button

Captura de Tela 2021-04-07 às 17 31 11

Login also fails when I try to run it. But its weird because I've got no error in any requests, nor any console logs

public/translations/cat-CAT.json
NeoVier: My bad, didn't even realize that was happening

src/elm/Session/LoggedIn.elm
NeoVier: Correct me if I'm wrong, but from what I understand, we always have the private key **on localStorage**, but not necessarily in the `Model`. The `Model` only has access to the private key when:
- The user just logged in (inserted their 12 words and created a pin)
- We prompt the user for their pin. This happens when the user refreshes the page, so the `Model` is reset, and would need to get the private key from local storage, and we need the pin for that (if I understood it correctly)

So this function is asking if **the Model** has the private key. And yes, we always have the auth token in the `Model` **and** localStorage

lucca65: right! you are completely right, totally forgot about that! my bad

src/elm/Session/LoggedIn.elm
NeoVier: I actually introduced a `type alias PrivateKey = String` later on and forgot to change stuff to use it (will do so). I don't think we need a `type PrivateKey` for this though, it would just mean more wrapping/unwrapping, and I can't see many benefits we get from that

src/elm/Session/LoggedIn.elm
NeoVier: Yeah, was having some issues where the `Feedback` would show up above the profile nav, I guess this makes sure that doesn't happen

src/index.js
NeoVier: Yep, this was pretty frustrating, but oh well :cry:

src/elm/Session/LoggedIn.elm
lucca65: i can see only one:

having an easy way to validate private keys from elm. that's something we don't do today, we only check if the length for the word list is correct. but there are actually reserved words and forgotten characters

in any way my suggestion was just to keep this opaque types pattern we've been using

NeoVier: I went to search for something, and found [this type](https://github.com/cambiatus/frontend/blob/f5c52eb84b6dd32cbc09ee7efa012a7199e79645/src/elm/Eos/Account.elm#L173-L174) that already exists. Should we use that?

Just to be sure, by private key you mean the [hashed analogue of the passphrase](https://github.com/cambiatus/frontend/blob/f5c52eb84b6dd32cbc09ee7efa012a7199e79645/src/elm/Auth.elm#L19-L21)?

Does it make sense to validate either the passphrase (12 words) (other than the word count) or the private key? They're both generated automatically, not from the user :thinking:

public/translations/en-US.json
NeoVier: Honestly, I don't even think it's possible to get to this state, but it was there before and I couldn't figure out how to test it, so I just kept it, but yes, we can do something like that.

Also, does it still make sense to get multiple `account_name`s for the same public key/user (there was a `LoginWithMultipleAccounts` state before that I removed, because it wasn't being generated)? Can we be sure that the correct account (if there is one) will always be at the first position in the list?

src/elm/Session/LoggedIn.elm
lucca65: > I went to search for something, and found this type that already exists. Should we use that?
I think so yeah, in any case we can have a different function that only takes model, this way you don't need to wrap, unwrap every time!

> Just to be sure, by private key you mean the hashed analogue of the passphrase?

> Does it make sense to validate either the passphrase (12 words) (other than the word count) or the private key? They're both generated automatically, not from the user 🤔

Actually the 12 words is a mnemonic that serves as base (with other information to generate the private key)

public/translations/en-US.json
NeoVier: Yes, this case is actually impossible.
For it to happen, the user must be logged in (have their info on localStorage), but this port is only called from the `Login` page (which means the user is not logged in). But, **if** this ever happens, I thinks it's safe to just ignore it: it would be the same as the user logging out and logging back in, with a different account (we can make sure all of the data a logged in user can have in localStorage is deleted every time on login, just to be safe)

src/elm/Session/LoggedIn.elm
NeoVier: > Actually the 12 words is a mnemonic that serves as base (with other information to generate the private key)

Ok, but does it make sense to validate any of these? We're dealing with

- 12 words (generated for the user, doesn't really make sense to validate)
- Private key (generated for the user, doesn't really make sense to validate)
- PIN (just a sequence of digits, not much to validate other than length and being only digits)

The only data we get from the user is from registration (name, username, email, etc), but I haven't gone through that in this PR

NeoVier:
@lucca65 I suspect your issue with pasting might be related to [permissions](https://web.dev/async-clipboard/#security-and-permissions). I've added some handling for that, so it's clearer to the user what's going on.
I'm not sure about not being able to log in, maybe it's somehow related to the clipboard again? :thinking: 

Anyway, I can't test on Safari/Mac, so can you try again please?

src/elm/Session/LoggedIn.elm
lucca65: > 12 words (generated for the user, doesn't really make sense to validate)
although its generated for they user have to type it on the text area, or paste it to login.

> Private key (generated for the user, doesn't really make sense to validate)
+1

> PIN (just a sequence of digits, not much to validate other than length and being only digits)
+1 Pin validation that we already have is enough

lucca65 (approved):
Still don't work on safari for iOS and have this strange behavior on Safari for Mac.

NeoVier:
Huh, that's weird. Nothing shows up as `Feedback` or on the console?

By the way, did you mean to approve this?

lucca65:
yes! i approved the pr

lucca65:
something must be up with netlify again

NeoVier:
Oh yeah, auth isn't working on netlify deploys - we still don't have `GRAPHQL_SECRET` defined, so it sends an empty string as the password and fails authentication. I've deployed this on staging.

@heltonlr can you test this out on staging?
It's mainly the login page flow. There's a button to paste from clipboard (when the browser supports it. **note: you need to [use https](https://staging.cambiatus.io) for this to work/show up**), better error handling, etc. Also try doing stuff that needs you to be authenticated and that asks for your PIN

heltonlr (commented):
Wasn't possible to login. This error appears
@NeoVier 

I am using Google Chrome on desktop, I created an account and tried to login with the 12 words.

Screen Shot 2021-04-10 at 09 44 20

heltonlr (approved):

PR #504, opened 2021-04-02
Feature/claims improvements 2

lucca65:
N/A

- Improves the criteria to know which claim is votable or not, by taking into consideration their deadline and usages
- Allow API sorting on the dashboard analysis query
- Improve text size on Claim page
- Show new status of "voting blocked" on the Claim page to be consistent with the action status
- Show tag of completed/closed action directly on the claim page to improve transparency for the validators

- Fiddle with the features described above! :)


NeoVier (changes requested):
Looks pretty good, I just think you forgot to include some translations and an extra unnecessary comment. As a side note, I think we can only use `Shared`'s time, and use that everywhere, but maybe that will be easier to do once the support subdomains PR hits, since it includes that broadcasting feature, we'd just need to request that it reloads it whenever we need it

NeoVier (approved):
Sweet :ok_hand:

PR #523, opened 2021-05-04
Enhancement/Phone numbers

### LINKED ISSUE #508: Support more phone numbers
With new communities and users coming into the platform, we need to support phone numbers from more countries

Support phone numbers from the following countries:
- [x] Portugal
- [x] Canada
- [x] United Kingdom
- [x] New Zealand

- Support specific countries (what we're doing now) with custom placeholders and icons
- Support all countries. We'd still have placeholders, but not well formatted (`11912345678` vs `11 91234 5678`), and would probably need to use a third-party icon set for the flags

**Additional context**
Came from [this slack thread](https://cambiatus.slack.com/archives/CA83HJAAD/p1619743728000300)

### LINKED ISSUE #510: Its not possible to delete contact information
There is no button to delete contact information, even with the API supporting it

A delete/thrash button somewhere.

N/A

**Additional context**
N/A

NeoVier:
Closes #508, closes #510

- Support new phone numbers:
  - [x] Portugal
  - [x] Canada
  - [x] United Kingdom
  - [x] New Zealand
- Add a better way to add new countries - just add them to `supportedCountries` in `Profile/Contact.elm`!
- Add "delete contact" button
- Let user delete contacts by just submitting empty ones in the profile edit contact page

Test the contact modal:
1. Log in to an account that doesn't have any contact information
2. Go to the dashboard
3. See the contact modal
4. Check if the country selector (when you select either "Phone" or "Whatsapp") behaves as expected and has all the countries

Test the profile page:
1. Go to your profile
2. Go to the contact edit page
3. Check if the country selector behaves as expected and has all the countries
4. Try deleting contacts with the contact button and by leaving them blank


lucca65 (changes requested):
it took me a second to realize that I have to hit save after deleting and confirming, why don't we simply delete it after the confirmation modal?

NeoVier:
> it took me a second to realize that I have to hit save after deleting and confirming, why don't we simply delete it after the confirmation modal?

+1, that really is confusing. I'm fixing this

> Also the dropdown only shows 4, I think its safe to show everything, at least keep the min height is the limit to the bottom o f the screen

IMO it looks too long if we show the entire thing (and will become longer and longer with new supported countries):
2021-05-04-130511_1308x433_scrot

And it kind of breaks the modal:
2021-05-04-130632_882x723_scrot

lucca65:
i see, my suggestion would be a bad solution then... what do you think about adding a transparent 50% white bg with a down arrow? showing that there is more below?

NeoVier:
Is the browser scroll bar not enough? :thinking:

NeoVier:
I just realized it fits **exactly** 4 countries, maybe we can change the height a bit, so it kind of crops another country? That would show there's more below

lucca65:
lots of browsers nowadays don't show scrolls bars like old windows versions where they are always visible and in this case must take 20~30% of the space the flags take, instead its a bar that appears on top and is only visible while scrolling

NeoVier:
> I just realized it fits exactly 4 countries, maybe we can change the height a bit, so it kind of crops another country? That would show there's more below

That's what that would look like:
2021-05-04-165541_1039x298_scrot

What do you think?

lucca65:
not beautiful but for sure solves the problem

lucca65:
When I have no contact it doesn't show the contact modal, did we reach that date limit?

lucca65:
adding and then removing telegram made the modal appear 🤔

lucca65 (approved):

NeoVier:
That might have been from a bug I had introduced earlier, but is now fixed. It made it so when you submitted empty contacts it would submit something like `https://t.me/`

PR #545, opened 2021-05-27
Fix/back button

### LINKED ISSUE #520: The back doesn't actually go back
**Describe the bug**
The app contains several different "back" buttons but none of them actually go back, instead they are hard coded to go to a certain page. This makes stuff confusing 

**To Reproduce**
A simple example:

1. Go to the community settings page of a community you are admin of
2. Click on "objectives and actions"
2. Click on "back"
3. See error

**Expected behavior**
Simply go back to same page as it would if we pressed the browser's back button

**Screenshots**
N/A

All

NeoVier:
Closes #520 

Make the back button always go to the previous page the user was in. If there is no previous page, just go to the dashboard

Go to pages that have the back button and see if it redirects you to the correct page. Make sure you don't fall into a loop. Here is an example navigation that could cause a loop, but shouldn't:

1. Go to `/dashboard`
2. Click on the cog icon on the top right - you'll be redirected to `/community/settings`
3. Click on the "Edit" button under "Objectives and Actions" - you'll be redirected to `/community/objectives`
4. Click the back button - you'll be redirected back to `/community/settings`
5. Click the back button again - you should be redirected back to `/dashboard`

### DIFF
--- a/src/elm/Page.elm
+++ b/src/elm/Page.elm
@@ -255,12 +255,14 @@ viewTitle text_ =
 
 
 viewHeader : LoggedIn.Model -> String -> Route -> Html msg
-viewHeader { shared, previousRoute } title route =
+viewHeader { shared, routeHistory } title route =
     div [ class "w-full h-16 flex px-4 items-center bg-indigo-500" ]
         [ div [ class "flex container mx-auto" ]
             [ a
                 [ class "flex items-center mr-4"
-                , previousRoute
+                , routeHistory
+                    |> List.drop 1
+                    |> List.head
                     |> Maybe.withDefault Route.Dashboard
                     |> Route.href
                 ]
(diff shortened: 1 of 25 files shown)


src/elm/Page.elm
```diff
@@ -254,13 +254,17 @@ viewTitle text_ =
...
     div [ class "w-full h-16 flex px-4 items-center bg-indigo-500" ]
         [ div [ class "flex container mx-auto" ]
             [ a
                 [ class "flex items-center mr-4"
-                , Route.href route
+                , routeHistory
+                    |> List.drop 1
+                    |> List.head
+                    |> Maybe.withDefault Route.Dashboard
+                    |> Route.href
```
lucca65, line 267: ohhhhh I never figured out how we would know this, dumb me it's simply a list 🤡

lucca65 (approved):
🔥🔥🔥🔥🔥

src/elm/Page.elm
NeoVier: To be fair, it took some trial and error to get it right and avoid any loops :laughing:

PR #546, opened 2021-05-28
Fix community's website button and cover photo

NeoVier:
Closes N/A

- Automatically include `https://` to community's website if they don't provide it
- Change default cover photo to use SVGs
- Show actions as completed on the claim cards (the ones in the analysis page and in the dashboard)

- Edit a community to have a website, but don't include `https://` in front of it. If you now log out and go to the "Join" page, you should be able to click the "Visit the website" button and be redirected to the correct website
- Go to a community that doesn't have a cover photo (e.g. `cambiatus`), and see if the right one shows up (note that they're slightly different imgs for desktop and mobile - the mobile one is wider)
- Look for a claim from a completed action that was still pending, and see if it shows as "Completed" instead of "Under Review"

## Additional context
- URL issue reported on Slack by @muguika for the Ticolones community on demo
- Change cover photo to SVGs requested by @rafachadud  on Slack
- Completed action status requested by @juramos-2020 on Slack


src/elm/Page/Community/New.elm
```diff
@@ -559,7 +559,12 @@ validateModel shared accountName model =
                         , hasObjectives = True
                         , hasKyc = False
                         , hasAutoInvite = model.hasAutoInvite
-                        , website = model.website
+                        , website =
+                            if String.startsWith "https://" model.website || String.startsWith "http://" model.website then
+                                model.website
+
+                            else
+                                "https://" ++ model.website
```
lucca65, line 567: i'm not sure about this, you see not all websites have https. good websites will automatically redirect to https:// when accessed on http

we do that!

lucca65 (changes requested):

src/elm/Page/Community/New.elm
NeoVier: If we don't add the protocl (`https://`/`http://`) at the start, we get [the issue Karla talked about in Slack](https://cambiatus.slack.com/archives/CA83HJAAD/p1622157075032400), where the browser understands the url as an internal url, so it redirects to something like `ticolones.demo.cambiatus.io/www.ticolones.com` instead of redirecting to the external url `www.ticolones.com`. We could change it in the href instead, but it's kind of the same thing anyway, and we'd need to do it everywhere else we might use the community's website

NeoVier: So the website doesn't even have the chance to redirect to `https://`, because the user isn't even redirected to it

NeoVier: Oh, ok :sweat_smile:

NeoVier:
@lucca65, @juramos-2020 asked for some prints from some features, and one of them was to show the action as completed in the claim cards (the ones in the analysis page and in the dashboard), and I realized that wasn't a thing! It's a quick fix, so I just put it in this PR as well, since it's filled with tiny fixes that should be live before the update :sweat_smile:

lucca65 (approved):
💣 🔥 !!

PR #561, opened 2021-06-09
Fix/cookie env

NeoVier:
Closes N/A

- Change cookie keys depending on environment (e.g. `bespiral.user` becomes `bespiral.user.staging` on staging, `bespiral.user.demo` on demo and stays as `bespiral.user` on prod)
- Fix bug with Profile.Summary where users couldn't scroll the page when hovering over the profile picture on desktop

Before this PR, there was a bug where users who logged on prod would see an error when they'd go to staging or demo. The steps to reproduce that error:

1. Clear all cookies from demo/staging and prod
2. Log into prod
3. Go to demo/staging, see error

Now, if you do this, you shouldn't see any errors. **Note: because cookies behave differently on the netlify deploy, you should test this on staging (at the time of this writing, it's already deployed there)**

### DIFF
--- a/src/elm/View/Components.elm
+++ b/src/elm/View/Components.elm
@@ -2,7 +2,7 @@ module View.Components exposing
     ( loadingLogoAnimated, loadingLogoAnimatedFluid
     , dialogBubble
     , tooltip
-    , bgNoScroll
+    , bgNoScroll, PreventScroll(..)
     )
 
 {-| This module exports some simple components that don't need to manage any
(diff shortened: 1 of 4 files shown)


src/elm/View/Components.elm
```diff
@@ -94,6 +94,24 @@ tooltip { t } tooltipMessage =
...
+    node "bg-no-scroll"
+        (attribute "elm-prevent-scroll-class" preventScrollClass
+            :: attrs
+        )
+        []
+
+
+type PreventScroll
+    = PreventScrollOnMobile
+    | PreventScrollAlways
```
lucca65, line 117: nice!

lucca65 (approved):

PR #578, opened 2021-06-19
Fix detection for when user is in demo

NeoVier:
Closes N/A

- How we detect if we're on demo

This is deployed on demo, so you can test it like so:
1. If you're already logged in on demo, log out
2. Try to go to a community that doesn't exist (e.g. https://thisisntacommunity.demo.cambiatus.io)
3. Be redirected to https://www.cambiatus.com/welcome-demo

### DIFF
--- a/src/elm/Session/Guest.elm
+++ b/src/elm/Session/Guest.elm
@@ -333,7 +333,7 @@ update msg ({ shared } as model) =
                 { currentUrl | host = "cambiatus.staging.localhost" }
                     |> Url.toString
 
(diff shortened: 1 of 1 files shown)


lucca65 (approved):
🤔

NeoVier:
`Url.toString` (what we were using before) returns the entire url (e.g. https://cmbx.demo.cambiatus.io/dashboard), so it didn't work with `String.endsWith`. `.host` returns just the host (e.g. `cmbx.demo.cambiatus.io`), that's why it works. Maybe we should have something on `Shared.Model` that tells us if we're on `Localhost | Staging | Demo | Production` :thinking:

PR #580, opened 2021-06-21
Enhancement/about community page

### LINKED ISSUE #540: Show Token data on the community page
**What we have**
Currently we show lots of information on the community page: members, transfers, claims, orders, products. 
Captura de Tela 2021-05-24 às 16 02 21

**Proposal**
We could include more info related to the token that the community uses. Here is a sample with some data: 
https://www.figma.com/file/268fiH2aIOgBMSKYybffuJ/About?node-id=0%3A1

We can add even more info!
Captura de Tela 2021-05-24 às 16 06 00
(we can ask for directions from the design team if needed)

**Why**
Communities already asked us about this information, also we could do a little improvement on those stats CSS/ look and feel

**How**
The implementation is simple enough, we can do a direct call to the blockchain to query that data. Here's a `cleos` example:

```sh
cleos -u https://muda.cambiatus.io get table cambiatus.tk MUDA stat
```

**Additional context**
N/A

NeoVier:
Closes #540 

- Update the section with data about the community on `/community`
- Improve `Float` formatting

1. Log into a community
2. On the dashboard, click on "How to earn <symbol>" (or just go directly to `/community`)
3. You should see the section with data about the community. Make sure it's looking like [the designs](https://www.figma.com/file/268fiH2aIOgBMSKYybffuJ/About?node-id=0%3A1)


src/elm/Eos.elm
```diff
@@ -255,29 +256,7 @@ symbolToString (Symbol symbol precision) =
...
-            else
-                withoutSeparator ++ separator ++ String.repeat precision "0"
-
-        beforeSeparator :: afterSeparator :: _ ->
-            beforeSeparator
-                ++ separator
-                ++ (afterSeparator
-                        ++ String.repeat (precision - String.length afterSeparator) "0"
-                        |> String.left precision
-                   )
```
lucca65, line 280: 🍠  SWEET

lucca65 (approved):
dammmn it looks 🔥 Everyone will love it! the spacing and harmony is much better!

Also, nice work refactoring the float formatting, amazing!

juramos-2020:
I can't see the section with data about the same community is in the design

NeoVier:
@juramos-2020 please use the netlify deploy to test this one :sweat_smile: Here's the link: https://deploy-preview-580--angry-wing-8fcdd0.netlify.app

juramos-2020:
Thank @NeoVier 😃

Now it's pending only card of the "Minimum balance"

NeoVier:
@juramos-2020 actually, the minimum balance card is implemented, but the Cambiatus community doesn't have a minimum balance, so it doesn't show up. If you try it on a community that does have a minimum balance, it should show up

@lucca65 is this intended? Should we default to showing a card with minimum balance == 0? :thinking:

juramos-2020:
@NeoVier 

accessing this link I only have access to two communities: cambiatus and Mizu. In both  it's pending only card of the "Minimum balance"

https://deploy-preview-580--angry-wing-8fcdd0.netlify.app

juramos-2020:
If the desired combat is from the card of the "Minimum balance" does not appear when the community does not have the "mininum balance", then everything is ok

juramos-2020:
I tested it with a community that had the minimum balance and it worked

juramos-2020:
Congratulations guys. It looks great 😉👍

NeoVier:
> @lucca65 is this intended? Should we default to showing a card with minimum balance == 0?

ping @lucca65

lucca65:
@NeoVier this is weird bro, it does have a min balance on staging, check it out:

Captura de Tela 2021-06-23 às 15 21 04

I'm checking the `stat` table, using the symbol_code (without precision) to check this information.... 🤔

NeoVier:
Huh... We actually get this info from Graphql, so it's probably some inconsistency between the blockchain and our database, or some bug on the backend

If we query it on Graphql we get `null`:

2021-06-23-152643_1254x227_scrot


I wonder if it should even be allowed to be `null`. It's something every community should have well defined, right?

lucca65:
oh... you are getting it from the GraphQL... I see. Must be something on event-source then

lucca65:
It can actually be null because of how we structure data on cambiatus. Cambiatus consists of two contracts:

1. Community: responsible for networking (invitations), objectives and Shop
2. Token: responsible for token dynamics and types, transfers, issues, retires etc

We can have they both independently, although the community contract becomes stagnant without a token, little can be done outside of defining params network. But on our database we only have one table for simplicity. Usually token stuff we query the blockchain directly because since its money, we utmost need the latest information. You'll notice that for balance we don't have a GraphQL API. I think the same must be applied here, to avoid problems of having wrong data because event-source is still catching up with it!

NeoVier:
Ok, got it. So the answer is we should query the blockchain directly for the `minBalance`?

lucca65:
yeah, and all the other token related stuff! Sorry bro, I'm kinda used to all that information, my faulty brain assumes everyone also knows this 🤦

src/elm/Api.elm
```diff
@@ -106,3 +107,27 @@ getExpiryOpts shared symbol toMsg =
...
+        , body =
+            Eos.encodeTableQuery query
+                |> Http.jsonBody
+        , expect =
+            Decode.field "rows"
+                (Decode.index 0
+                    (Decode.field "supply" Eos.decodeAsset)
+                )
+                |> Http.expectJson toMsg
+        }
```
lucca65, line 133: for some reason my mind thought you were already getting the minBalance from this function 🙃

NeoVier:
No problem! I wonder if it makes sense to have the token and community stuff separate... I think it would be less confusing that way, and this kind of thing wouldn't happen anymore. What do you think?

Anyway I think for this PR we should just fetch the minBalance from the blockchain, and separate the token and community data on another issue, as that will probably require some changes on the app (if you think separating them is a good idea)

lucca65:
> No problem! I wonder if it makes sense to have the token and community stuff separate... I think it would be less confusing that way, and this kind of thing wouldn't happen anymore. What do you think?

Yeah, as we talked when you were working on the new community admin screens, I do think it could be separate, our database don't even need to hold those info at all and our frontend could have them as different stuff. We need to adjust some stuff but conceptually cambiatus could still working with several different types of tokens with that structure...

Anyway, +1 to making them separate

lucca65 (approved):
:shipit:

NeoVier:
@lucca65 I've done some very shallow improvements towards separating community and token here, just for code organization. However, in order to not fill this PR with a bunch of unrelated stuff, created #583 so we can tackle this more specifically there

PR #582, opened 2021-06-23
Enhancement/add tests

### LINKED ISSUE #535: Use automated tests
**What we have**
Right now we test everything manually, which makes it easy to let some edge cases slip in to the final build. Our development process goes like this right now:

1. Produce code
2. Test locally
3. Fix code if needed
4. Get a review from a teammate, and fix the code if needed. Repeat this until the PR is approved
5. Ask one of the designers to test the PR. If they request changes, go back to 4.
6. Once a dev and a designer have approved the PR, be happy and merge it.

**Proposal**
We should start using automated tests

**Why**
This would improve the overall quality of our app, and ensure we don't break anything between updates. We also would require less time from the design team to test things, which means they gain time and we gain time, win-win!

**How**
Probably using [elm-test](https://github.com/elm-explorations/test) initially. In the future, we could look into tools like [Selenium](https://www.selenium.dev/) to do some more advanced tests.

**Additional context**
N/A

NeoVier:
Closes #535 

- Layout initial testing structure
- Add `elm-test` to our `pre-push` hook
- Add `elm-test` to our CI/CD pipeline on github
- Create examples of how to test with `elm-test`
- Create some helper functions for testing
- Define a baseline for what tests should look like and how they should be organized in our project
- Introduce some tests for already-existing modules/functions

Note: This PR might not cover every single thing that needs testing right now. It's meant to cover some things, and establish examples for future us to use as a reference. This PR adds many tests, but only because we didn't have any tests, so this should be an exception! In order to maintain tests up to date, pull requests after this one should introduce tests alongside their enhancements/features/bug-fixes

Well, we can't write tests for tests :sweat_smile:, so look over the overall structure and organization, check if you agree with it or if we could improve on something. Check if the tests make sense, cover the functions appropriately, and actually pass. Tests should be run on github alongside our other checks, like `elm-review` and build. The actual code of the app shouldn't change (too much), which means we're not really risking breaking the app with this PR


lucca65 (commented):
Nice! I like it how you are focusing on testing functions with business rules instead of screen flows! +1

tests/Session/SharedTests.elm
```diff
@@ -0,0 +1,185 @@
...
+            , port_ = Nothing
+            , path = "/"
+            , query = Nothing
+            , fragment = Nothing
+            }
+
+        makeInput url =
+            { url = url, environment = Flags.Production }
+    in
+    describe "when environment is production"
```
lucca65, line 44: nice!!

tests/TestHelpers/Fuzz.elm
```diff
@@ -0,0 +1,130 @@
+module TestHelpers.Fuzz exposing
```
lucca65, line 1: i'm curious, where the name fuzz comes from? i've seen "factory" "mocker" "faker" etc, but never fuzz 🤔

maybe i didn't catch the ideia but this is the same concept as those other names right?

lucca65 (approved):
🧫

tests/TestHelpers/Fuzz.elm
NeoVier: I have two answers for you:

Why I chose this name:
Basically to follow the names from the [test package](https://package.elm-lang.org/packages/elm-explorations/test/latest/). You can see there's `TestHelpers.Random`, `TestHelpers.Shrink` and `TestHelpers.Fuzz`, which then are imported with the aliases `Random`, `Shrink` and `Fuzz` respectively, which are the same names you would use if you were to use the test package directly. These modules just provide `Generator`s, `Shrinker`s and `Fuzzer`s that are specific to our project.

Why it's named like this on the test package:
Probably because of [Fuzzing](https://en.wikipedia.org/wiki/Fuzzing) (or fuzz tests), as described in the [Test.fuzz documentation](https://package.elm-lang.org/packages/elm-explorations/test/latest/Test#fuzz), which also mentions other ways this kind of test is called.

(By the way, there's a pretty interesting text about shrinkers [here](https://package.elm-lang.org/packages/elm-explorations/test/latest/Shrink#what-are-shrinkers-and-why-do-we-need-them-), explaining what they are and how they can help us)

PR #587, opened 2021-07-08
Enhancement/transfers

### LINKED ISSUE #204: Some date time information is shown incorrectly
**Describe the bug**
Date time is shown incorrectly in actions/claims/verifications. On our versions we only show date (not time), so it was difficult to notice. 

When trying to reproduce this error please take into consideration that our server is in UTC, it will depend on the time of the day you are, the time when the claim was opened, your timezone, etc.

**To Reproduce**

*Note*: I'm assuming the following scenario:
Server timezone: UTC
My computer timezone: GMT-3 (no daylight saving time)
Claim opened 22:00 GMT-3

1. Create similar or equivalent scenario as the one mentioned above
2. Open a claim for an action with at least one account you access as validator
3. Open the validator screen (the one you, as a validator, vote)
4. See error, the date will be displayed as the claim opened in the next day

**Expected behavior**
Show the real date the claim was opened

**Screenshots**
I wasn't able to get any screen 

Muda (muda.cambiatus.io)

### LINKED ISSUE #579: Improve transfer list
**What we have**
A very old, not community aware transfer list. We need to update that

**Proposal**
Improve our transfer view to follow the new design: https://www.figma.com/file/47BEsGEMSWfhTm3zC0ugyA/Ticolones-version?node-id=1%3A421

**Why**
As an effort to improve visibility and as a preparation for the community admin balance tool

**How**
N/A

### LINKED ISSUE #581: Implement new transfers on dashboard
Implement the new transfer interface on the dashboard. It should use the new transfer query (https://github.com/cambiatus/backend/issues/176)

NeoVier:
Closes #581, closes #579, closes #204 (that's an old one!!)

- Use the new transfer API from cambiatus/backend#176
- Use new design for the transfers list on the Dashboard
- Introduce a localized date viewer with support to show stuff like "Today", "Yesterday", and any other custom string (if we choose to), along with a localized general date, like "Claimed on 7 june 2021". For this we needed a couple of things:
  - A new `Language` type for better organization
  - A custom element to display the formatted date (so we can use the nice JS api)
  - `Shared` now holds a `Time.Zone` (which solves #204)
- Fix some issues with `Profile.Summary` in a couple of pages
- Add a linter for our CSS. This way we have linters for all languages we use (JS, CSS and Elm), and have a nice standard
- Introduce a new customElement, `infiniteList`, which is used to have things like feeds from social media, where more data is requested automatically as the user scrolls
- Roll back the hotfix for v1.11.1

- Test according to the description #204 
- Make sure the transfer list shows up on the dashboard, and you can access a specific transfer (either by notifications, or by clicking on one on the dashboard)
- See if the dashboard design is according to the figma link in #579 
- Check if you see anything weird with regards to dates (such as on claims, transfers, etc.)


lucca65:
🔥 🔥 🔥 🔥 🔥 

damn bro! this looks incredible, its not even finished and already is in a [r/nextfuckinglevel](https://www.reddit.com/r/nextfuckinglevel/)

src/styles/elm-datepicker.css
```diff
@@ -134,3 +134,11 @@
...
   color: white;
 }
+
+.elm-datepicker--container.relative-table .elm-datepicker--picker {
+  @apply relative;
+}
+
+.elm-datepicker--container.relative-table .elm-datepicker--table {
+  @apply w-full;
+}
```
lucca65, line 144: line ending missing

lucca65: maybe we can add a linter rule for this?

NeoVier: I'll look into a css linter we can use. I think the other linters should already report that :thinking:

src/elm/Claim.elm
```diff
@@ -381,24 +380,31 @@ viewClaimCard loggedIn profileSummaries claim =
...
             loggedIn.shared.translators
 
-        ( claimStatus, textColor ) =
-            case claim.status of
-                Approved ->
-                    ( t "all_analysis.approved", "text-green" )
+        -- date dateTime =
+        --     Just dateTime
+        --         |> Utils.posixDateTime
+        --         |> Strftime.format "%d %b %Y" Time.utc
```
lucca65, line 386: solitary commented function

lucca65 (changes requested):
I don't like the way it looks on the desktop, I know we are mobile first, but it just don't look right, the font size of the title is different, the spacing is too big... idk folks, can we improve it a little bit? 😀 
Captura de Tela 2021-07-19 às 17 20 57

NeoVier:
It does look off without the claim cards... Here's how it looks when you have claims to vote on (I've adjusted the font weight on "My timeline" to match with the "Actions for analysis" text):

image

I think it looks better with them (the text matches with "Actions for analysis", there's more things that go further to the left), but I guess most users don't see them anyway... Maybe we should change the size/weight of the "Welcome to [community]" text to fit with the other ones?

Here's how it looks if the "Welcome to [community]" text has the same size/weight as the other ones:

image

Not sure what we could do about the spacing... Any ideas @juramos-2020 @rafachadud ?

lucca65:
+1 @NeoVier 

Looks like the timeline just takes so much more attention because it don't follow the grey background, besides the fonts... idk, I love it but its just takes so much attention

juramos-2020:
@lucca65 @NeoVier 

Here is the design of the new transfer. I believe this proposal solves the problem of excess white space.

https://www.figma.com/file/lbiu1Kx4wnDMs0azaR4eOh/My-timeline-Transfers?node-id=10%3A3018

Some points:
* on the desktop version as the scroll has I removed the "show more" button due to limited space. If the scroll can be added to the mobile version it will be great.

* About the color: 
It would not be good for us to link the color red to our transfer list, red refers to an error, something negative. And because it is a vibrant color, it ends up drawing more attention than the other elements. We don't want to convey that feeling to our users.

So I did a quick search on other financial products apps and noticed that they keep a default color for all types of transfers and highlight only those receiving GREEN.

As our green is very light (a study to change this hue is already on the design radar 😅) we chose the color blue (#45469B) to represent this transaction.

What do you think?

NeoVier:
Wow, I like the use of space a lot better! I'm just not sure about the gray used for transfers that the user sent... it looks like it's disabled, just feels a little weird. Maybe we could use something closer to black? I don't know if that would be too much contrast though :grimacing:.

lucca65:
+1! I love the way it is right now with red and green the way it was. 

But echoing @NeoVier, this new design is 🔥

juramos-2020:
Hum. I had done a test with black but it really drew a lot of attention.

I did a test now with this color next to black. What do you think?

We would have to include this color in our Design System. But I think that would not be a problem (I think lol)

https://www.figma.com/file/lbiu1Kx4wnDMs0azaR4eOh/My-timeline-Transfers?node-id=10%3A3018

NeoVier:
It's not a problem at all! And it looks much better like that, for me that's the one!! 🔥🔥

lucca65:
@NeoVier maybe we can identify the end of scroll to load more results? 👀 it would also look 🔥

NeoVier:
That definitely sounds interesting :eyes:... Looks like a job for custom elements yet again, and I'm pretty sure we can do a generic element without needing a `update`/`subscription`/`Model`/anything other than a view function. That way we can reuse it in other areas of the app in the future

lucca65 (approved):

juramos-2020:
Everything is ok with the transfer timeline. 👌😍👍

Just this "my offers" card that was left out of alignment. I would have to align this button with the button on the side card.

Teste_timeline

NeoVier:
As this PR is taking a long time, and that fix is super simple, I'll add that into #605 when I merge with master, that way we don't need to wait for another approval. Thanks for the feedback @juramos-2020!

PR #601, opened 2021-07-19
Remove char max len. in action description & photo instruction

### LINKED ISSUE #592: Remove 256 characters limitation from Actions form
**What we have**
Today we have 256 characters for the action description and on the validation with photo description

**Proposal**
Simply remove the limitation on those two fields

**Why**
The blockchain already clips the bigger part, and event-source can save it on Postgres normally

**How**
Just remove it!

**Additional context**
Discovered during our Full Team Session on 13 july

victorgdev:
Closes #592 

Remove the character limit of 256 to accommodate the changes on the backend for the following fields:
- [x] [action] photo_proof_instructions
- [x] [action] description
- [x] [community] description
- [x] [objective] description
- [x] [product] description

Also changed the description display behavior for objective descriptions. When displayed in ```/community``` page extensive obj descriptions truncate when not toggled, and are fully displayed when toggled. 

**CREATED** 
- Community
- Action
- Objective
- Offer

**As** comm admin
**Go to** ```action or objective or community or offer``` editor
**Click on** on edit description
**See** action description and photo instruction does not have char limits

**NEW**
- Community
- Action
- Objective
- Offer

**Go to** create button
**Fill out** form
**See** description field does not have char limits

### DIFF
--- a/src/elm/Page/Community.elm
+++ b/src/elm/Page/Community.elm
@@ -243,6 +243,29 @@ viewObjective loggedIn model metadata index objective =
             else
                 " "
 
+        objDescCharCount =
+            String.length objective.description
+
+        objDivStyle =
+            let
+                objMainClass =
+                    "px-3 py-4 bg-indigo-500 flex flex-col sm:flex-row sm:items-center cursor-pointer"
+            in
+            if isOpen && objDescCharCount > 40 then
+                class (objMainClass ++ "sm:h-auto")
+
+            else
+                class (objMainClass ++ "sm:h-10")
+
+        objTextStyle =
+            class
+                (if isOpen then
+                    "text-white font-medium text-sm sm:flex-grow-8 sm:leading-normal sm:text-lg"
+
+                 else
+                    "truncate overflow-hidden whitespace-nowrap text-white font-medium text-sm overflow-hidden sm:flex-grow-8 sm:leading-normal sm:text-lg"
+                )
+
         isDisabled =
             case loggedIn.claimingAction.status of
                 Action.ClaimInProgress _ _ ->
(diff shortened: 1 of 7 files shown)


victorgdev:
@lucca65 and @NeoVier, per our last conversation how should we proceed with this proposed change? 
- Should we create a higher limit as oppose to simply remove it? 
- Also remembering that the DB is set with 256 varchar, we should probably create an issue for the backend migration to hold the defined change, correct?

lucca65:
I'll address this change on the backend later today @victorgdev, I'll let you know!

lucca65:
Here is the issue: https://github.com/cambiatus/backend/issues/181

It also include a description of all fields changed, it should be changed on the frontend as well

lucca65:
done and deployed on staging @victorgdev

NeoVier (changes requested):

src/elm/Page/Community.elm
```diff
@@ -293,7 +273,8 @@ viewObjective loggedIn model metadata index objective =
     else
         div [ class "my-2" ]
             [ div
-                [ objDivStyle
+                [ class "px-3 py-4 bg-indigo-500 flex flex-col sm:flex-row sm:items-center cursor-pointer"
+                , classList [ ( "sm:h-auto", isOpen && objectiveDescriptionCharCount > 40 ), ( "sm:h-10", not isOpen || objectiveDescriptionCharCount < 40 ) ]
```
NeoVier, line 277: Hmm, maybe we should use padding instead of a set height for this (or at least adjust the padding so it's consistent when it's open and when it's closed). It's very difficult to detect whether the text should wrap or not based on the length of the text (every character has a different width), so some objectives might be 40 characters long, but they might not wrap. Like this:
2021-08-04-193131_1269x546_scrot

Notice how the open objective gets taller than the others, but it doesn't wrap. Also, when you open something that does wrap, it's still kind of weird that the text goes down a bit.

Here's one closed:
2021-08-04-193238_1212x98_scrot

And the same one, but open:
2021-08-04-193248_1214x451_scrot

Notice how when it's open, there's a bigger gap between the text and the top of the div.

NeoVier (approved):
:man_student:

juramos-2020:
Hi @victorgdev ok, it's working 😉👍

PR #605, opened 2021-07-20
Enhancement/better sentry errors

### LINKED ISSUE #543: Show balance when transferring
Its really annoying on our monthly claims, I have to recheck my balance every time 

Just show the balance that we already have on the transfer form

maybe it can be a modal, but I think the problem would persist even if we do it 🤷

### LINKED ISSUE #552: Improve sentry error reporting
**What we have**
We log a lot of errors to sentry that aren't relevant, which makes it so actual relevant issues get buried below these, making it harder to know if there's actually something wrong with the app or not. At the time of this writing, out of the top 6 most recurring events, 4 are related to DOM manipulation, with over 125k combined events.

**Proposal**
Solve or suppress the most recurring issues, depending on whether it's worth it to fix them or not

**Why**
If we only have relevant issues, we can do a better job debugging with sentry, and actually use it correctly. It'll make our lives way easier in regards to checking the health of our app

**How**
N/A, depends on each issue

**Additional context**
Brought up by @lucca65 on Slack

NeoVier:
Closes #552 . I consider this as closing, but it doesn't *solve* any of the current issues on Sentry. However, some of them won't be shown anymore, and the ones that are shown, will have better descriptions and should (hopefully) be easier to debug.

- Update the sentry package.
- Rethink the `Log` module. Before this PR, the `Log` module was supposed to be generic. Now, it's specific to Sentry, which lets us have way clearer and more specific error messages, details, tags, etc. We now **always** include the place in the code where the error happens (module name and function name), a message, a list of tags, and a list of extra information. Since the location in the code isn't super granular, it's encouraged to not have duplicate error messages. We also always send the Msg that triggered the event. Another addition is having breadcrumbs in Elm! This is fantastic because we can define steps of what happened before the event. The API was thought out so it could be extensible, with nice helper functions that define basic stuff, but also possible to use it at a lower-level to define more details.
- Use the new `Log` module all over the app (which is why there are a lot of changes).
- Include a new elm-review rule that ensures we always have a correct location when sending events (ensures that module name and function name are correct). **Note**: This rule is limited in the sense it won't check records that use variables instead of literal `String`s on the record, so always prefer doing something like
```elm
{ moduleName = "Main", function = "update" }
```
over
```elm
{ moduleName = variableWithModuleName, function = variableWithFunctionName }
```
- Remove some old and unused code about push preferences.
- Add South Africa as a supported country.
- Check what is the maximum value a user can transfer.
- Show user balance when transferring. (Closes #543)

This will also change everything about how we report errors to sentry, so we will have some old and stale ones, we might have to filter them after a while after deploying this PR.

You have a couple of options:
1. Clone this branch, run `yarn install && yarn start` and watch the console while you use the app. On development, there will be stuff printed out to the console starting with `[====BREADCRUMB]` or `[====EVENT]`. Events are sent to Sentry right when they happen, while breadcrumbs are sent with events. (Events and breadcrumbs are only sent to Sentry on production (staging/demo/communities). On your local machine, they will just be printed to the console)
2. Run the deployed version of the app, try to cause some error and watch [the issues in Sentry](https://sentry.io/organizations/cambiatus/issues/?project=1480468).

There is also [this](https://sentry.io/organizations/cambiatus/issues/2523169823/?project=1480468&query=is%3Aunresolved) example event that was generated from Elm for testing purposes if you want to look at it (this shouldn't happen anymore though, it was just a hardcoded event)


lucca65 (commented):
Wow, this looks already so much better, specially the breadcrumbs!

Bro, I've simulated an issue in the netlify env: https://sentry.io/organizations/cambiatus/issues/2523636281/?project=1480468&query=is%3Aunresolved

Here is what I done: tried to transfer an absurd amount with lots of zeros to someone 10000000!! Some stuff I've noticed!

1. we never actually show the error on the form
2. We got a blockchain error, and that much is clear, but we cannot know the params used, nor the actual blockchain error. 
3. We still got a weird type error, maybe its something related to the JS part?

NeoVier:
Is there a maximum value a person is allowed to transfer at once?

> We still got a weird type error, maybe its something related to the JS part?

It is... JS and `undefined` :roll_eyes:

lucca65:
> Is there a maximum value a person is allowed to transfer at once?

Yeah, its limited by the available balance + community limit (how negative you can be)

NeoVier:
> we never actually show the error on the form

Done!

> We got a blockchain error, and that much is clear, but we cannot know the params used, nor the actual blockchain error.

The error was actually fine, but when we were reporting the eos error, we got another error (the one with `undefined`), and the eos error didn't go through. Here's how it looks now: https://sentry.io/organizations/cambiatus/issues/2524750360/?project=1480468&query=is%3Aunresolved (you won't be able to reproduce this error because we validate the amount before sending the transaction, but that's how it would look)

lucca65:
> Done! Also solved #543 while doing that

Wowwwww! It looks so much better now!!!!

lucca65 (changes requested):
Oh I see! Another error got in the way first, fair!

Do you think we could send JSON data into the breadcrumb as well? I wonder if we could do this for two reasons:

1. With this we have improved context on what the user did before reaching an error. like this error that does not provide us with the transaction JSON (although the error message itself and the breadcrumbs is enough in this case) https://sentry.io/organizations/cambiatus/issues/2524750360/?project=1480468&query=is%3Aunresolved
2. and most exciting, we could use Sentry to track app usage with alerts, getting to know when we got more errors of a certain kind

I mean this isn't blocking, but could be cool for future us

Captura de Tela 2021-07-22 às 14 34 17

NeoVier:
I'm not sure I understood what you meant... We can (and do) send JSON on breadcrumbs as well as on events.

In that case, `logEvent.start` is a generic breadcrumb we add when we start logging any event, so it's not really practical to include specific details of the event itself there. On development, the entire data sent is logged, but since it can possibly contain sensitive information, that's not sent to Sentry.

If you mean adding a JSON object inside the breadcrumb/event info, it seems like we can't. However, a solution is to stringify the value if it's an array or object. So, in this case, the information is related to the event (as opposed to the breadcrumb), and can be found on the "EOS TRANSACTION" part, check it out:

Before:
2021-07-22-165855_1304x418_scrot

Now:
2021-07-22-165918_1314x405_scrot

It's not optimal, but now we can see the `Sent` data. (Here's the link if you want to take a deeper look: https://sentry.io/organizations/cambiatus/issues/2524750360/?project=1480468&query=is%3Aunresolved)

lucca65:
check this error out, this is a prod error, and we can actually use JSON properly: https://sentry.io/organizations/cambiatus/issues/2052029671/?project=1480468&query=is%3Aunresolved&statsPeriod=14d

maybe this is a different field?

lucca65:
I guess text discussion over this is getting a little bit confusing 😁 maybe we can use our meeting tomorrow to discuss this properly (and in ptbr!)

NeoVier:
Ok, not sure what I did, but JSON works now :man_shrugging:. But yeah, we can talk better tomorrow!

NeoVier:
@lucca65 here's what I've done:

1. About logging objects: that problem seemed to only happen on netlify deploys. But anyway, I've rolled back the version of Sentry we were using, and it seems to work now.
2. About the name of eos transactions: we now get specific messages from the error object, and prepend them with `[EOS]` (we also can always search for them with the `cambiatus.type:eos-transaction` tag).

lucca65 (approved):
🔮

PR #610, opened 2021-07-27
Fix after deploy v1.11

NeoVier:
Closes N/A. Fixes some issues after deploying v1.11

Removed queries that weren't deployed on the backend at the time of the update.

- Check if the shop still works when you're logged in. It should no longer work as a guest :(
- Check if the payment history (`/payments/:user`) page works fine


lucca65 (approved):
kinda sad, let's try to make the next version release to be as soon as possible!

PR #625, opened 2021-08-19
Enhancement/new typography

### LINKED ISSUE #622: Atualizar a fonte e tipologia do app de acordo com o novo padrão
**What we have**
Usamos um determinado tamanho de fonte, contraste e espaçamento

**Proposal**
Melhorar a legibilidade e uso geral do app para as novas diretrizes [especificadas aqui](https://www.figma.com/file/rT5DMvPAtvbHSxNBwQENDM/New-shop---v1?node-id=0%3A1

**Why**
Enquanto esperamos atualizações do backend para o novo shop, podemos adiantar essas melhorias, que na realidade afetam todo o app 

**How**

**Additional context**

NeoVier:
Partially addresses #622 

- Use new font sizes
- Adjust some styling
- Add `Html.label` and `Html.textarea` to the list of forbidden functions outside of certain modules

Go through the app and see if it all looks right


juramos-2020:
It seems to me that everything is ok 😉👍

NeoVier:
Yesterday, during our dev meeting, me and @lucca65 were talking about how different the app looks with the new font sizes. We think it could be due to two things:
1. We're just so used to seeing it a certain way, that when we have this big of a change, it just looks weird.
2. There are sizes that are just too big. Before this change, we had more font sizes, and they weren't so different from each other. Now, we have really big jumps from body to heading to h3, h2, h1. The differences are about 8px for each of them. Prior to this, we had smaller increments (of about 2px each!).

Also, it might be because I used the wrong font sizes on certain elements. I tried to keep them around the same "level" as before, but might have missed some of them.

What do you think about this @juramos-2020 @heltonlr? Does the app look how you envisioned it? @lucca65 do you have anything else to add? Maybe you could share those other examples you were talking about during the meeting?

juramos-2020:
Thank you Henrique.  Let's review all screens 😉👍

heltonlr (approved):

NeoVier:
As this is going to be (already is) a long running PR/change, and it's going to affect the entire app, I'll create a feature branch, and change this PR to point to that new branch. @heltonlr is doing an amazing work of reviewing every screen on the app, and we can do them screen by screen or flow by flow. This PR added the initial structure for the new typography stuff and also reviewed the login/register/join flow. We already have a new Dashboard on the horizon, and that'll be already in another PR to the same feature branch. This way it'll be easier to manage everything and keep organized

@lucca65 do you want to review this branch, or can I just merge it to the feature branch?

lucca65:
i think we can merge into the feature branch and move from there, i guess to will be better to evaluate the final product

PR #628, opened 2021-08-19
Enhancement/pin hidden

### LINKED ISSUE #589: Make the PIN as hidden by default
**What we have**
Today the default of the pin is to be shown, the user can toggle this to change to hidden

**Proposal**
Default the value to hidden

**Why**
To improve privacy

**How**
N/A

**Additional context**
Discovered during our Full Team session

NeoVier:
Closes #589 

- Default the PIN to hidden
- Use localStorage when the `USE_SUBDOMAIN` flag is set to `false`. This way we can actually have some persistence, especially on Netlify deploys
- When the user submits their PIN, save (on localStorage/cookies) if it should be hidden or shown on the next time their PIN is requested

1. Be logged out
2. When you enter your PIN, it should be hidden. If you change it to visible, the next time you enter your pin anywhere on the app, it should be visible, etc. That should work with any PIN request.

Your PIN may be requested on login, when you're changing your PIN on your profile page, or when you're signing blockchain transactions (such as approving a claim, buying a product, transferring to a friend, etc.). Note that if you're logged in, we remember your PIN from the first time you enter it, and don't ask for it again until you refresh the page.

### DIFF
--- a/src/elm/Auth.elm
+++ b/src/elm/Auth.elm
@@ -56,8 +56,8 @@ import View.Pin as Pin
 -- INIT
 
 
-init : Maybe Eos.PrivateKey -> Model
-init maybePrivateKey_ =
+init : Bool -> Maybe Eos.PrivateKey -> Model
+init pinVisibility maybePrivateKey_ =
     let
         status =
             case maybePrivateKey_ of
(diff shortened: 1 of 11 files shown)


lucca65 (approved):
Looking nice! 🏗️

PR #629, opened 2021-08-23
Fix mouseEnter/Leave handler on profile.summary to be fired only on img hovering

### LINKED ISSUE #593: Show Profile.Summary only when hovering a user's photo
**What we have**
The Profile.Summary popup (on desktop) shows up when hovering the `div` that holds the user's profile photo and name

**Proposal**
Only show the popup when hovering the user's profile photo directly

**Why**
The popup can be annoying sometimes and it opens too easily

**How**
Move the event that triggers the popup opening

**Additional context**
Requested during our Full-Team Session

victorgdev:
Closes #593  

Fix mouseEnter/Leave handler on profile.summary to be fired only when hovering the avatar image of user

Go to any claim card
Find user avatar
Hover with mouse

### DIFF
--- a/src/elm/Profile/Summary.elm
+++ b/src/elm/Profile/Summary.elm
@@ -113,7 +113,8 @@ view shared loggedInAccount profile model =
 mobileView : Shared -> Eos.Name -> Profile.Basic profile -> Model -> Html Msg
 mobileView shared loggedInAccount profile model =
     div [ class "md:hidden cursor-auto" ]
-        [ viewUserImg shared loggedInAccount profile True model
+        [ viewUserImg profile True model
+        , viewUserNameTag shared loggedInAccount profile model
         , Modal.initWith { closeMsg = ClosedInfo, isVisible = model.isExpanded }
             |> Modal.withBody
                 [ div [ class "pt-14" ]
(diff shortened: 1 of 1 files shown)


tailwind.config.js
```diff
@@ -173,6 +173,22 @@ module.exports = {
...
+          },
+          '100%': {
+            'opacity': '1',
+            'pointer-events': 'auto'
+          }
+        }
+      },
+      animation: {
+        fadeIn:
+          'appear 50ms linear 400ms both'
```
NeoVier, line 191: Usually tailwind rules are kebab-cased. I'd say it's nice to follow their convention. Also, we could put the string on the same line as the key
```suggestion
        'fade-in': 'appear 50ms linear 400ms both'
```

tailwind.config.js
```diff
@@ -173,6 +173,22 @@ module.exports = {
...
+      keyframes: {
+        appear: {
+          '0%': {
+            'opacity': '0',
+            'pointer-events': 'none'
+          },
+          '100%': {
+            'opacity': '1',
+            'pointer-events': 'auto'
+          }
```
NeoVier, line 186: When the key is a single word, with no special characters, we don't need to add it as a string (note the `opacity` key right above, on line 173)
```suggestion
        appear: {
          '0%': {
            opacity: '0',
            'pointer-events': 'none'
          },
          '100%': {
            opacity: '1',
            'pointer-events': 'auto'
          }
```

NeoVier (dismissed):
Just some code-style changes. Sorry I didn't see these on the pair session, the screen was really blurry :grimacing: :eyeglasses:

src/elm/Profile/Summary.elm
```diff
@@ -150,7 +163,7 @@ viewUserImg shared loggedInAccount profile isMobile model =
             [ container [] [ Avatar.view profile.avatar model.imageSize ]
             , if not isMobile && model.isExpanded then
                 View.Components.dialogBubble
-                    { class_ = "w-120"
+                    { class_ = "w-120 animate-fadeIn opacity-0"
```
NeoVier, line 166: After changing the tailwind config, we need to change it here too, to use `animate-fade-in` (right now, the popup doesn't even show)
```suggestion
                    { class_ = "w-120 animate-fade-in opacity-0"
```

NeoVier (changes requested):

PR #631, opened 2021-08-26
Enhancement/modal

### LINKED ISSUE #618: Close modal when pressing Esc
**What we have**
We already close the modal if the user clicks out of it.

**Proposal**
Close the modal when the user presses Esc.

**Why**
Better accessibility, it's kind of expected to happen. Also, it would mean we wouldn't need to listen for all key presses on LoggedIn, making for less spamming on the debug console.

**How**
Listen for key presses, figure out if it's Esc, and call the close Msg.

**Additional context**
N/A

NeoVier:
Closes #618 

- Fix overflowing issue with the "Add contact" modal
- Make the contact flag selection element and the user nav a lot more accessible
- Add a `keyListener` custom element. It's useful to listen for keys without the need for a subscription. It's very extensible, allowing us to listen to any event, such as `keydown`, `keypress` and `keyup`. It also allows for multiple (typed) keys to be listened to, and to `stopPropagation`. We use this element to listen for `Escape` on modals.
- Add a `focusTrap` custom element. It acts like a `div`, but won't let focus get outside of it. [We should be careful when using it though](https://developers.google.com/web/fundamentals/accessibility/focus/using-tabindex#modals_and_keyboard_traps)

Testing modals:
1. Open up a modal
2. Press `Tab` a bunch of times. That should cycle the focus within the modal, but shouldn't focus elements outside of the modal. Do the same with `Shift+Tab` (to go in the opposite direction)
3. Make sure you can exit it by pressing `Esc` on your keyboard, clicking outside of it and by clicking on the "X" button on the top

Testing the contact flag selection:
1. Go to either the "Edit contact" page on your profile, or to the "Add contact" modal on the dashboard (it only shows up if you have no contact options)
2. Press `Tab` and notice how the select is only focused once. If you enter the select, you should see that pressing the up key or down key cycles through the options

### DIFF
--- a/src/elm/Profile/Contact.elm
+++ b/src/elm/Profile/Contact.elm
@@ -17,6 +17,7 @@ module Profile.Contact exposing
     )
 
 import Api.Graphql
+import Browser.Dom
 import Cambiatus.Enum.ContactType as ContactType exposing (ContactType(..))
 import Cambiatus.Mutation
 import Cambiatus.Object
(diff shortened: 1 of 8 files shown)


juramos-2020:
Hello guys

Testing modals: - it's working 👌
Testing the contact flag selection: -  it's working 👌

It was great 👌

PR #636, opened 2021-09-03
Enhancement/responsive search

### LINKED ISSUE #617: More responsive search
**What we have**
We only get search results when we press Enter.

**Proposal**
Get search results while we are typing, like the [tailwind docs](https://tailwindcss.com/).

**Why**
Better UX, it would feel faster than having to press enter.

**How**
Do the same thing as pressing Enter, but while the user is typing.

**Additional context**
Suggested by @juramos-2020 on #608

NeoVier:
Closes #617 

- Get recent searches only when community is done loading (there were apparently some race conditions before)
- Adjust maximum number of recent searches (was 3, now it's 10)
- Improve accessibility (with tabindex and focus styles)
- Close search when pressing Esc
- Request search results as the user types

1. Open the search bar
2. Make sure you can close it by pressing Esc.
3. Open the search bar again
4. Start typing something. Results should show up as you type. If you erase it, you should see your recent searches again.
5. Try navigating with Tab. It should be clear what element you're focusing, and it should follow a logical order (left-to-right, top-to-bottom)

### DIFF
--- a/src/elm/Search.elm
+++ b/src/elm/Search.elm
@@ -26,7 +26,7 @@ import Graphql.Http
 import Graphql.OptionalArgument exposing (OptionalArgument(..))
 import Graphql.SelectionSet as SelectionSet exposing (SelectionSet, with)
 import Html exposing (Html, a, br, button, div, h3, img, li, p, span, strong, text, ul)
-import Html.Attributes exposing (autocomplete, class, classList, minlength, required, src)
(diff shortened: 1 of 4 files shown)


lucca65 (approved):
Nice! Looks like it should be like that since always! Such a big improvement

juramos-2020:
hI @NeoVier 

IT'S working for me 😉👍

PR #693, opened 2022-02-23
Feature/update notification preferences

### LINKED ISSUE #681: Allow users to update their notification preferences
Since the implementation of our News feature, every month users will receive a monthly digest of all that happened in the community. We would need to allow our users to toggle those preferences of receiving emails, otherwise we might be not be respecting their choice of receiving communication from us. It is a global config, you opt on receiving preferences from all communities or none of them, so it's important to make this clear on our interface.

One other preference that we need to setup, its language. We already have a menu to change the in app language, so we can hook this language change to also trigger a mutation.

We need to discuss how we are going to present this, but we already have a mutation to set those preferences in the profile:

```graphql
mutation {
  preference(claimNotification: false, digest: true, language: PTBR) {
    account
    name
  }
}
```

Right now we have these notification options:

- claim notification
- monthly digest about the community
- transfer notification

We also need to handle the case where the user has not set their language on the backend yet. It will come as `null` on the GraphQL `users` object. If that's the case then we need to set the current chosen language during the first initial setup of the app after its logged.

We could have a basic setting directly on the menu, maybe as a button that opens up

Captura de Tela 2022-01-28 às 11 08 40

NeoVier:
Closes #681 

- Add toggles to control notification preferences (transfers, claims and digest)
- Send user's preferred language on app startup, and whenever the user changes the language of the app

- Go to your profile page
- See if the toggles look right
- Play with them, and refresh the page to see if they kept their state (if you turned one of them off, it should remain off after refreshing the page)
- Make sure you can't see the toggles on profiles that are not yours


src/elm/Profile.elm
```diff
@@ -125,6 +130,32 @@ selectionSet =
         |> with User.analysisCount
         |> with (User.kyc Kyc.selectionSet)
         |> with (User.address Address.selectionSet)
+        |> with (User.claimNotification |> SelectionSet.map (Maybe.withDefault True))
+        |> with (User.digest |> SelectionSet.map (Maybe.withDefault True))
+        |> with (User.transferNotification |> SelectionSet.map (Maybe.withDefault True))
```
NeoVier, line 135: Same thing as above ☝️

lucca65 (commented):
Looks great! I'll go ahead and merge this

PR #695, opened 2022-02-24
Fix Form's NoOpt handling

### LINKED ISSUE #692: Fix error handling on forms
**Describe the bug**
There is a very hidden edge case where the error reporting for forms doesn't work. It has to do with `OptionalResult.OptNothing`. When a form has a specific combination of `OptNothing`s, the result of the form is always `OptNothing`, making it impossible to validate the form. One such combination is a `Form.arbitrary` inside a `Form.withGroup`, and can be seen on the `/community/new` form.

**To Reproduce**
1. Go to `/community/new`
2. Fill out the form
3. Click submit
4. See that nothing happens. If you use the Elm debugger, you'll see a `Form.NoOp` msg every time you click submit.

**Expected behavior**
The form should be validated correctly - if there is an error, show the error, or submit the form if there isn't an error.

**Screenshots**
N/A

staging/localhost

**Additional context**
N/A

NeoVier:
Closes #692 

- Introduce `Form.arbitraryWith`, which can be used instead of `arbitrary` when needed
- Fix form validation logic for `/community/new`
- Fix issue where users couldn't resubmit pin if they provided the wrong pin

- Go to `/community/new`
- Fill out the form
- Click "Create"
- See the form working (before this PR, it wasn't possible to create a new community, because of this bug)


lucca65:
Something looks weird but I'm not sure if this is related.

When I try to use the form, it only requires me to fill the symbol, and if I do, it allows me to submit the form, but simply reload it, no errors. Then if I simply resubmit it says "the URL is already being used" 🤷

NeoVier:
Validation for that form should be better now. Here are the rules:

- Description can be empty (I don't think there's a problem with that, but let me know if we should enforce some character limits)
- Currency name must have at least 2 characters
- Community subdomain must form a valid URL (we automatically insert `http://` if needed, and also `{{env}}.cambiatus.io`). We also now force it to lowercase automatically (there was an issue about that recently on some community)
- Symbol must form a valid symbol (same as before, but now we always show it as uppercase)
- All of the number fields remain the same (they only need to be valid floats)
- Community website remains the same (can be empty)
- Logo remains the same

Also fixed an issue where you couldn't try again if you submitted a wrong pin

lucca65:
Bro, the input removes the suffix when there is an error, may be confusing 

Captura de Tela 2022-03-03 às 13 12 08
Captura de Tela 2022-03-03 às 13 12 15

lucca65 (changes requested):
Bro, the input removes the suffix when there is an error, may be confusing 

Captura de Tela 2022-03-03 às 13 12 08
Captura de Tela 2022-03-03 às 13 12 15

lucca65:
The app is somewhat usable: I can go to profile and notifications but with no community selected its a limbo

NeoVier:
> Bro, the input removes the suffix when there is an error, may be confusing

Done ✅ 

---

I managed to replicate that other error, and as I suspected, it's because there was an error fetching the new community:
image

That's probably another case where cambiatus/backend#148 would be helpful...

I haven't run into this problem when running with `USE_SUBDOMAIN=true`, maybe because there's a delay between `creating the community > redirecting to new page > fetching all the JS stuff > requesting community` vs just `creating the community > requesting community`, but it's pretty much pure luck, and we can't guarantee it won't happen

So... I'm not really sure how to deal with this without listening for an event when the backend has received that data. I can only think of a couple of "solutions":

1. Forcefully wait for a couple of seconds before trying to fetch the new community, and hope the backend has synced up with the blockchain (not really great)
2. Just let the error happen and have a nicer error screen, telling the user what happened and possible solutions (reload the page, check internet connection, reach out to us, or reach out to the community (not very usable in this case, but failing to load the community could happen to anyone))
3. Leave it as is until we have cambiatus/backend#148

How do you think we should proceed?

lucca65:
maybe number 4:

- If form errors, show error
- if form succeed, simply return to dashboard with current community and show a feedback telling the user that it works but might take a while to show

Not ideal, but should be enough, community creation is not a common workflow

NeoVier:
Ok, done! I also found a bug where we weren't sending the correct subdomain, so I fixed that as well, on the `new` and `info` forms

lucca65 (approved):
Nice, you gave this form a good update bro! congratulations!!

PR #732, opened 2022-04-23
Fix profile PDF download and clipboard access on Safari

NeoVier:
Closes N/A. Raised by Verdes on Slack: https://cambiatus.slack.com/archives/G01GNQB5G5B/p1650728963382709

- Fix the PDF download from the profile
- Require permissions API to read from the clipboard. Specifically, Safari doesn't support the permissions API. [In Safari, to access the Clipboard API, it must come from a user event](https://stackoverflow.com/a/62566097) (like clicking a button). Apparently, if we call a port from elm, that doesn't count as a user event anymore (?), so we get an error that says we don't have permissions to do that. For now, we can just require that the permissions API is implemented in the user's browser. Other than that, all we can do is:
  - Wait for Safari to implement permissions API (it's listed as an Experimental Feature on Safari 15.4)
  - Use some deprecated way to access the clipboard.

Testing profile PDF:
- Go to your profile
- Try to download your 12 words PDF

Testing paste on Safari:
- Go to the login page on Safari
- See that the "Paste" button is no longer there

### DIFF
--- a/src/elm/Auth.elm
+++ b/src/elm/Auth.elm
@@ -51,8 +51,8 @@ import View.Pin as Pin
 -- INIT
 
 
-init : Bool -> Maybe Eos.PrivateKey -> ( Model, Cmd Msg )
-init pinVisibility maybePrivateKey_ =
+init : Maybe String -> Bool -> Maybe Eos.PrivateKey -> ( Model, Cmd Msg )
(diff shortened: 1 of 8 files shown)


lucca65 (approved):
🕺🏻

PR #736, opened 2022-04-27
Add cambiatus.version tag to Sentry

### LINKED ISSUE #733: Add version tag to Sentry errors
**What we have**
We use Sentry to log runtime errors on production.

**Proposal**
Some time ago, we started showing the current commit hash on the app to help us debug errors from users. However, we only have access to that if users send us screenshots or something like that. We could also send this commit hash to Sentry.

**Why**
We can eliminate "false" bugs, caused by browser caching and things like that. If we resolve a bug, we can probably archive it or delete it on Sentry so that we know it's fixed.

**How**
On the `logEvent` function in JS, add a tag that has the value of `process.env.COMMIT`

**Additional context**
I found a very weird error that shouldn't be able to happen: https://sentry.io/organizations/cambiatus/issues/3222078152/?project=1480468&referrer=slack

I'm pretty sure it's been a while since we've made changes to the `Invite`, so I don't think that error should happen, and the only thing I can think as the cause is a wrong version of the frontend. The only way to be sure of that kind of thing is if we had commit hashes on Sentry.

NeoVier:
Closes #733 

- Add a `cambiatus.version` tag to Sentry errors. The tag's value will be the current commit hash (the same one we show on the footer)

Checkout this error: https://sentry.io/organizations/cambiatus/issues/3229438253/?project=1480468&referrer=slack

image

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -119,6 +119,7 @@ const logEvent = (event) => {
       }
       // If the error comes from Elm, this key will be overwritten
       scope.setTag('cambiatus.language', 'javascript')
+      scope.setTag('cambiatus.version', process.env.COMMIT)
       scope.setTags(tags)
       scope.setTransaction(transaction)
       contexts.forEach((context) => {
--- a/src/index.js
+++ b/src/index.js
@@ -47,6 +47,8 @@ if (env !== 'development') {
       return unusedCategories.includes(breadcrumb.category) ? null : breadcrumb
     }
   })
+
+  Sentry.setTag('cambiatus.version', process.env.COMMIT)
 }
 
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -47,6 +47,8 @@ if (env !== 'development') {
       return unusedCategories.includes(breadcrumb.category) ? null : breadcrumb
     }
   })
+
+  Sentry.setTag('cambiatus.version', process.env.COMMIT)
```
lucca65, line 51: lol

lucca65 (approved):
Good bro! kkkk

PR #739, opened 2022-04-29
Use backend api manifest

NeoVier:
Closes N/A.

**We should wait for the backend PR to be merged before merging this**

- Change the manifest link to point to the manifest served by our backend after https://github.com/cambiatus/backend/issues/218

To make things easier, deploy the correct version of the backend and frontend on staging.

Then, go to a community (e.g. https://cambiatus.staging.cambiatus.io). The browser should load the correct information on the manifest 

Next, go to a url that doesn't represent a community (e.g. https://asdfasdfa.staging.cambiatus.io). The browser should follow the 301 redirect and serve the generic manifest.json, which is stored in the frontend.

Browsers have different ways to show what the manifest file looks like:
- Safari shows the request in the network tab, you can just see the response
(it differs from browser to browser on how to check this - it might show up in the network tab, or in the application tab of the devtools)
- Chrome and Firefox show the manifest in the application tab (and the request doesn't show up on the network tab 🤷)
- Other browsers may show it somewhere else, but it should be there!

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -7,8 +7,13 @@
     <!--
          manifest.json provides metadata used when your web app is added to the
          homescreen on Android. See https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/
+
+         We first try to get the manifest from our backend, which will populate
+         it with data related to the current community (based on the subdomain
+         the user is at). If it can't find a valid community, it redirects to the
+         manifest file we keep in the `public` folder
     -->
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
+    <link rel="manifest" href="%PUBLIC_URL%/api/manifest">
     <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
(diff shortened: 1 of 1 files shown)


MatheusBuss (commented):
@NeoVier I couldn't get this to work properly on firefox and chrome in staging. The route `/api/manifest` returns 404. Is this because of a mismatch between the frontend and backend used by staging? I assume so, because from our tests, this should be working.

NeoVier:
Oh yeah, I was testing some other things and had deployed another version of the backend.

I just deployed the correct versions now, please try again 🙇 

Also, this isn't supported by all browsers yet. You can check compatibility here: https://caniuse.com/web-app-manifest. It might be worth trying with unsupported browsers as well just to be sure though (browsers are weird 🤷 )

MatheusBuss:
> I just deployed the correct versions now, please try again 🙇

Working as intended 🙌

MatheusBuss (approved):

PR #827, opened 2022-09-02
Enhancement/elm book elm review

### LINKED ISSUE #809: Add elm-review to elm-book
**What we have**
We have a really nice elm-review config in our app. However, we're not using that config on our elm-book

**Proposal**
Use the same review config (or at least a similar one) on elm-book. We should also run it on our CI

**Why**
Because elm-review is awesome, and we should use it

**How**
If elm-review supports it, we could just use the same config, but point at a different elm project. Otherwise, maybe we could use UNIX symbolic links (we already do this for some files on elm-book)

**Additional context**
Discovered we're not running elm-review on elm-book during https://github.com/cambiatus/frontend/pull/805

NeoVier:
Closes #809 

- Add elm-review to elm-book
- Add elm-review on elm-book on our CI and pre-push hook
- Fix elm-review errors

- Check that elm-book is being reviewed by elm-review on our CI

### DIFF
--- a/.github/workflows/on-push-action.yml
+++ b/.github/workflows/on-push-action.yml
@@ -37,7 +37,9 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          working-directory: ./elm-book
+          elm_review_config: elm-book/review
+          elm_json: elm-book/elm.json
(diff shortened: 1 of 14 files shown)


lucca65 (commented):
🧹

lucca65 (approved):
🧹

